PR #2, opened 2019-03-31
better detection of runtime avro version

radai-rosenblatt:
Signed-off-by: radai-rosenblatt <radai.rosenblatt@gmail.com>


radai-rosenblatt:
no way to check for the existence of classes/methods without exception.
the upside is this happens once, on boot, so the amortized cost is ~0.
another option would be trying to get the version from the avro jar manifest, but thats more fragile as any sort of shading/fat-jaring would wipe the manifest (or worse)

PR #7, opened 2019-04-12
Avro FastSerde support

gaojieliu:
This code change is a fork of RTBHOURSE's avro-fastserde implementation.
https://github.com/RTBHOUSE/avro-fastserde
You could refer to the original repo to find more details about the original
idea for this initiative.
This code change introduces the following new features:
1. For deserialization:
   a. Back-port it to Avro-1.4;
   b. Object reuse;
   c. Special optimization for float vector;
   d. Align with regular avro implementation to return 'Utf8' for 'string' type;
2. For serialization:
   a. So far, it still doesn't support Avro-1.4;
   b. Align with regular avro implementation to support both 'String' and 'Utf8' for 'string' type;


avro-fastserde/src/main/java/com/linkedin/avro/fastserde/Utils.java
```diff
@@ -0,0 +1,49 @@
...
+  public static AvroVersion getRuntimeAvroVersion() {
+    return AvroCompatibilityHelper.getRuntimeAvroVersion();
+  }
+
+  public static List<AvroVersion> getAvroVersionsSupportedForDeserializer() {
+    return AVRO_VERSIONS_SUPPORTED_FOR_DESERIALIZER;
+  }
+
+  public static List<AvroVersion> getAvroVersionsSupportedForSerializer() {
+    return AVRO_VERSIONS_SUPPORTED_FOR_DESERIALIZER;
```
gaojieliu: Typo here, and I will change it to `AVRO_VERSIONS_SUPPORTED_FOR_SERIALIZER`.

avro-fastserde/src/test/java/com/linkedin/avro/fastserde/generated/avro/AnotherSubRecord.java
```diff
@@ -0,0 +1,282 @@
+/**
+ * Autogenerated by Avro
```
smccauliff: What's with all the autogenerated code being committed?

gaojieliu: What is the recommended way for those auto-generated classes in this repo?
Without those classes, 'gradle build` will fail right away.
I understand that we should trigger the command to generate those classes first then run 'gradle build', but it seems not so convenient.

FelixGV: I guess these are two build management philosophies.

Approach 1 is to check in the generated classes, such that if someone clones the repo and opens it up in their IDE, it works right away (i.e.: no red lines for classes missing). This is the approach Gaojie is following in this PR.

Approach 2 is to not check in the generated classes, such that if someone clones the repo, they have to run the build at least once, so that the build auto-generates the classes, prior to the project looking fine in the IDE.

Based on @smccauliff's comment, it seems like this project might prefer following approach 2. As long as there is a good way in Gradle to generate Avro classes for multiple Avro versions, then that is fine. I guess, worst case scenario, in the absence of a clean Gradle way, we could make Gradle shell out to the `regenerate_avro.sh` script... If we did that, we would probably want to add the generated classes to `.gitignore`.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -0,0 +1,529 @@
...
+   *
+   * @param writerSchema
+   *            {@link Schema} of written data
+   * @param readerSchema
+   *            {@link Schema} intended to be used during deserialization
+   * @return specific-class aware avro {@link FastDeserializer}
+   */
+  public FastDeserializer<?> getFastSpecificDeserializer(Schema writerSchema, Schema readerSchema) {
+    String schemaKey = getSchemaKey(writerSchema, readerSchema);
+    FastDeserializer<?> deserializer = fastSpecificRecordDeserializersCache.get(schemaKey);
```
FelixGV: I guess we do a `get()` followed by `putIfAbsent()` here instead of `computeIfAbsent()` because of JDK-8161372, right?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerializer.java
```diff
@@ -0,0 +1,10 @@
+package com.linkedin.avro.fastserde;
+
+import java.io.IOException;
+import org.apache.avro.io.Encoder;
+
+
+public interface FastSerializer<Type> {
```
FelixGV: Same as above: recommend calling the generic type `T` or `TYPE`.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerializerGeneratorException.java
```diff
@@ -0,0 +1,18 @@
+package com.linkedin.avro.fastserde;
+
+public class FastSerializerGeneratorException extends RuntimeException {
+
+  private static final long serialVersionUID = 1L;
```
FelixGV: We don't need this do we? Using Java Serialization is a well-known anti-pattern, so hopefully nothing here relies on that... I'd suggest just removing this.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
```diff
@@ -0,0 +1,291 @@
...
+import java.util.Collection;
+import java.util.Iterator;
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericArray;
+import org.apache.avro.generic.GenericData;
+import org.apache.avro.io.Decoder;
+
+
+/**
+ * This is a re-implementation of Avro's {@link GenericData.Array} class.
```
FelixGV: Should we leave a `TODO: Provide arrays for the other primitive types`?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
```diff
@@ -0,0 +1,291 @@
...
+
+  public PrimitiveFloatList(Collection<Float> c) {
+    if (c != null) {
+      elements = new float[c.size()];
+      addAll(c);
+    }
+  }
+
+  /**
+   * Instantiate (or re-use) and populate a {@link PrimitiveFloatList} from a {@link org.apache.avro.io.ResolvingDecoder}.
```
FelixGV: Any `Decoder` will do, and in fact, we stand to gain from NOT using a `ResolvingDecoder`. We should update the JavaDoc link accordingly.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
```diff
@@ -0,0 +1,291 @@
...
+
+  @Override
+  public Float get(int i) {
+    return getPrimitive(i);
+  }
+
+  /**
+   * Add a primitive float inside the list, without boxing.
+   */
+  private boolean addPrimitive(float o) {
```
FelixGV: According to an offline discussion with @radai-rosenblatt, he thinks we don't even need this, and that the simple act of storing references to primitive floats instead of object Floats would allow the JVM to optimize away the boxing/unboxing sequence. In any case, it makes the code extra explicit to keep it as such, and we haven't tested whether it's true that the JVM behaves that way, so we should probably keep it this way for now. Perhaps we can leave a TODO to verify if it's enough to simply store the data as primitives in order to benefit from the GC optimization.

avro-fastserde/src/test/java/com/linkedin/avro/fastserde/micro/benchmark/AvroGenericDeserializer.java
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.avro.generic.GenericDatumReader;
+import org.apache.avro.io.BinaryDecoder;
+import org.apache.avro.io.DatumReader;
+import org.apache.avro.io.DecoderFactory;
+
+
+public class AvroGenericDeserializer<V> {
+
+  /**
+   * This default is intended for the backend. The client overrides this, and its default is defined in
```
FelixGV: This comment is not relevant in this context.

avro-fastserde/src/test/java/com/linkedin/avro/fastserde/micro/benchmark/AvroGenericDeserializer.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericDatumReader;
+import org.apache.avro.io.BinaryDecoder;
+import org.apache.avro.io.DatumReader;
+import org.apache.avro.io.DecoderFactory;
+
+
+public class AvroGenericDeserializer<V> {
```
FelixGV: Why are we including this class? Only for use by the micro-benchmark? Is this necessary? Why not make the micro-benchmark use `DatumReader` directly instead?

avro-fastserde/src/test/java/com/linkedin/avro/fastserde/generated/avro/AnotherSubRecord.java
radai-rosenblatt: im working on better codegen support (See avro-codegen module) that would allow generating the java code out of avro schemas (via a utility 1st then a gradle plugin).

i believe once that exists, and incombination with the idea plugin (https://docs.gradle.org/current/userguide/idea_plugin.html) it would be possible to define a codegen goal and make the idea task depend on the codegen one.

IIUD that would make intellij run the codegen when opening the project, and a similar thing could be done for builds in general.

however, it may be too soon to do it now

avro-fastserde/regenerate_avro.sh
```diff
@@ -0,0 +1,96 @@
...
+if [ ! -f $DEFAULT_AVRO_TOOLS_182_JAR ]; then
+  wget "https://repo1.maven.org/maven2/org/apache/avro/avro-tools/1.8.2/avro-tools-1.8.2.jar" -P avro_tools/avro_tools_182
+fi
+
+AVRO_SCHEMAS_PATH=(
+  "src/test/resources/avro/fastserdetest.avsc"
+  "src/test/resources/avro/defaultsTest.avsc"
+  "src/test/resources/avro/stringableTest.avsc"
+)
+CODE_GEN_PATH=(
```
radai-rosenblatt: could we add a comment explaining what the intended purpose of this structure is? at 1st glance it looks like a copy-paste error

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGeneratorBase.java
```diff
@@ -0,0 +1,303 @@
...
+import javax.tools.JavaCompiler;
+import javax.tools.ToolProvider;
+import org.apache.avro.Schema;
+import org.apache.avro.io.parsing.Symbol;
+import org.apache.log4j.Logger;
+
+
+public abstract class FastDeserializerGeneratorBase<T> {
+  public static final String GENERATED_PACKAGE_NAME = "com.linkedin.avro.fastserde.deserialization.generated";
+  public static final String GENERATED_SOURCES_PATH = "/com/linkedin/avro/fastserde/deserialization/generated/";
```
radai-rosenblatt: File.pathSeparator? (i realize this code is far from being windows-compatible, but still)

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGeneratorBase.java
```diff
@@ -0,0 +1,303 @@
...
+  private String compileClassPath;
+
+  FastDeserializerGeneratorBase(Schema writer, Schema reader, File destination, ClassLoader classLoader,
+      String compileClassPath) {
+    this.writer = writer;
+    this.reader = reader;
+    this.destination = destination;
+    this.classLoader = classLoader;
+    this.compileClassPath = compileClassPath;
+    codeModel = new JCodeModel();
```
radai-rosenblatt: +"this." for consistency :-)

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGeneratorBase.java
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    return printName;
+  }
+
+  public static int getSchemaId(Schema schema) {
+    Integer schemaId = SCHEMA_IDS_CACHE.get(schema);
+    if (schemaId == null) {
+      /**
+       * SchemaNormalization.toParsingForm(schema) is not available in avro-1.4
```
radai-rosenblatt: we've back-ported it - see com.linkedin.avro.compatibility.SchemaNormalization.toParsingForm and AvroCompatibilityHelper.toParsingForm()

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGeneratorBase.java
```diff
@@ -0,0 +1,303 @@
...
+      printName = (String) field.get(symbol);
+      field.setAccessible(false);
+    } catch (ReflectiveOperationException e) {
+      throw new FastDeserializerGeneratorException(e);
+    }
+
+    return printName;
+  }
+
+  public static int getSchemaId(Schema schema) {
```
radai-rosenblatt: i think "getSchemaFingerprint" is a better name for what this does

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGeneratorBase.java
```diff
@@ -0,0 +1,303 @@
...
+    } else if (Schema.Type.ARRAY.equals(readerSchemaType)) {
+      return "Array" + description + "Deserializer" + writerSchemaId + "_" + readerSchemaId;
+    } else if (Schema.Type.MAP.equals(readerSchemaType)) {
+      return "Map" + description + "Deserializer" + writerSchemaId + "_" + readerSchemaId;
+    }
+    throw new FastDeserializerGeneratorException("Unsupported return type: " + readerSchema.getType());
+  }
+
+  protected static String getVariableName(String name) {
+    return name + nextRandomInt();
```
radai-rosenblatt: i think a static final AtomicInteger + getAndIncrement() would lead to more readable (maybe even consistent) results and is safer

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGeneratorBase.java
```diff
@@ -0,0 +1,303 @@
...
+  }
+
+  protected static String getSymbolPrintName(Symbol symbol) {
+    String printName;
+    try {
+      Field field = symbol.getClass().getDeclaredField("printName");
+
+      field.setAccessible(true);
+      printName = (String) field.get(symbol);
+      field.setAccessible(false);
```
radai-rosenblatt: hide the evidence :-)

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -0,0 +1,1061 @@
...
+      deserializerClass = classPackage._class(className);
+
+      JVar readerSchemaVar = deserializerClass.field(JMod.PRIVATE | JMod.FINAL, Schema.class, "readerSchema");
+      JMethod constructor = deserializerClass.constructor(JMod.PUBLIC);
+      JVar constructorParam = constructor.param(Schema.class, "readerSchema");
+      constructor.body().assign(JExpr.refthis(readerSchemaVar.name()), constructorParam);
+
+      Schema aliasedWriterSchema = writer;
+      /**
+       * {@link Schema.applyAliases} is not working well in avro-1.4 since there is a bug in this function:
```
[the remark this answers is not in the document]
radai-rosenblatt: this sounds like something we can port/copy into the compatibility helper class (looks like a bunch of static logic we could easily make work under 1.4 as well)

avro-fastserde/src/test/java/com/linkedin/avro/fastserde/generated/avro/AnotherSubRecord.java
gaojieliu: Removed the generated classes from the commit

avro-fastserde/build.gradle
```diff
@@ -0,0 +1,139 @@
...
+
+configurations {
+  avro14
+  avro17
+  avro18
+}
+
+test {
+  //we run test multiple times with different avro
+  //versions, so dont use the default test task
```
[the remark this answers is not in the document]
gaojieliu: Sounds good.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializer.java
```diff
@@ -0,0 +1,14 @@
+package com.linkedin.avro.fastserde;
+
+import java.io.IOException;
+import org.apache.avro.io.Decoder;
+
+
+public interface FastDeserializer<Type> {
```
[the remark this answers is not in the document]
gaojieliu: Changed it to use `T`.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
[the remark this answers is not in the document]
gaojieliu: Updated the comment as suggested.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -0,0 +1,1061 @@
...
+        processFixed(schema, methodBody, action, putExpressionIntoParent, reuseSupplier);
+        break;
+      default:
+        processPrimitive(schema, methodBody, action, putExpressionIntoParent, reuseSupplier);
+        break;
+    }
+  }
+
+  private void ifCodeGen(JBlock parentBody, JExpression condition, Consumer<JBlock> thenClosure,
+      Optional<Consumer<JBlock>> elseClosure) {
```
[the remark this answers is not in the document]
gaojieliu: Sure.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -0,0 +1,1061 @@
...
+            if (Utils.isAvro14()) {
+              newFixedExpr = JExpr._new(fixedClass).arg(fixedBytesArray);
+            } else {
+              newFixedExpr = JExpr._new(fixedClass).arg(getSchemaExpr(schema)).arg(fixedBytesArray);
+            }
+            return newFixedExpr;
+          } else {
+            // fixed implementation in avro-1.4
+            // The specific fixed type only has a constructor with empty param
+            // The following may not work if the specific fixed code is generated through 1.7
```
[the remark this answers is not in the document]
gaojieliu: Removed the invalid comment since the test is running the fixed type with generic/specific against Avro 1.4/1.7/1.8.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
gaojieliu: Sure.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
gaojieliu: Done.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
gaojieliu: Sure.

avro-fastserde/src/test/java/com/linkedin/avro/fastserde/micro/benchmark/AvroGenericDeserializer.java
gaojieliu: Removed.

avro-fastserde/src/test/java/com/linkedin/avro/fastserde/micro/benchmark/AvroGenericDeserializer.java
gaojieliu: Those are for micro-benchmark.
Considering the benchmark could use all the functionalities, such as serialize single-object/multiple objects, and de-serialize single object/multiple objects w/o reuse, and I would like to keep those two utilities to make the benchmark test simpler.

avro-fastserde/regenerate_avro.sh
gaojieliu: Left some comments to explain the purpose, and let me know if they are not good enough.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGeneratorBase.java
gaojieliu: Adopted the recommendations!

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGeneratorBase.java
gaojieliu: Sure.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGeneratorBase.java
gaojieliu: The new commit started using this utility!

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGeneratorBase.java
gaojieliu: Sure.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGeneratorBase.java
gaojieliu: You meant `File.seperator`, right?

gaojieliu:
> what is the difference between src/test/resources/avro and src/test/resources/schema ?

```
 src/test/resources/avro
```
is used to store the avro schemas to compile, and the other folder is used to store the schemas, which will be directly used by the tests.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
gaojieliu: Probably since this is the original implementation..

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerializer.java
gaojieliu: Will change it to `T`.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerializerGeneratorException.java
gaojieliu: Will remove it.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGeneratorBase.java
FelixGV: The fingerprint concept already exists within modern versions of Avro though. Would this cause confusion if there are two kinds of fingerprints (ours and Avro's) ?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -0,0 +1,1059 @@
...
+        ifCodeGen(methodBody,
+            // Check whether schema matches
+            reuseVar.ne(JExpr._null()).
+                cand(reuseVar._instanceof(indexedRecordClass)).
+                cand(
+                    // The following statement will produce the equality check by "==", so that the comparison is
+                    // not trying to compare the schema content, but just the reference, which should be very fast.
+                    JExpr.invoke(JExpr.cast(indexedRecordClass, reuseVar), "getSchema").eq(recordSchema)),
+            thenBlock -> thenBlock.assign(result, JExpr.cast(indexedRecordClass, reuseVar)),
+            elseBlock -> elseBlock.assign(result, finalNewRecordInvocation)
```
FelixGV: Looks great! Thanks for doing the refactoring!

PR #12, opened 2019-10-09
Add new APIs to AvroCompatibilityHelper

ahmedahamid:
Add the following APIs to `AvroCompatibilityHelper`:
  - `newInstance(Class, Schema)`:    Instantiates objects of a specific type 
  - `getSchema(Type)`: 		  	   Locates the Avro schema associated with a Java (SpecificRecord) type
  - `getDefaultValue(Schema.Field)`: Retrieves the default value of an Avro field, if any

### DIFF
--- a/avro-migration-helper/src/main/java/com/linkedin/avro/compatibility/AvroCompatibilityHelper.java
+++ b/avro-migration-helper/src/main/java/com/linkedin/avro/compatibility/AvroCompatibilityHelper.java
@@ -12,6 +12,7 @@ import java.io.ObjectInput;
 import java.io.ObjectOutput;
 import java.io.OutputStream;
 import java.lang.reflect.Modifier;
+import java.lang.reflect.Type;
 import java.util.Collection;
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericData;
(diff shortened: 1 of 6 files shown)


avro-migration-helper/src/main/java/com/linkedin/avro/compatibility/AvroCompatibilityHelper.java
```diff
@@ -104,6 +109,33 @@ public static JsonDecoder newJsonDecoder(Schema schema, String input) throws IOE
     return FACTORY.newFixedField(ofType, contents);
   }
 
+  /**
+   * Create an instance of a class. If the class implements {@link SchemaConstructable},
+   * call a constructor with a {@link Schema} parameter, otherwise use a no-arg constructor.
+   * @param clazz class type to instantiate
+   * @param schema Avro schema to use for instantiating {@param clazz} if it implements {@link SchemaConstructable}
+   */
+  public static Object newInstance(Class clazz, Schema schema) {
```
radai-rosenblatt: can this be constrained to IndexedRecord ?

ahmedahamid: It can but this would be too restrictive since it would exclude `Fixed` for instance. I chose to preserve the signature of [`SpecificData.newInstance()`](https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/specific/SpecificData.java#L466)

PR #13, opened 2019-10-09
Use curl instead of wget for better portability

ahmedahamid:
Tested on mac and RHEL7.


FelixGV:
Good catch... wget is not there by default on mac. Thanks! Let's merge it in.

PR #14, opened 2019-11-12
Dynamically infer classpath by classloader according to the class usage

gaojieliu:
When application is using play framework with hot-reload enabled, fast-avro
couldn't find the right classpath by extracting from system property:java.class.path.
So this code change will explicitly find all the classes being used in the
generated files and put the corresponding libraries in the compile classpath.
Play framework is using both current classloader and the thread context classloader
to discover class path in hot reload feature. In fast-avro class generation/compilation,
we have to do the same time to align with this behavior.
Once the generated file is compiled successfully, loading the compiled classes
is not a problem in play framework with hot-reload enabled.

### DIFF
--- a/avro-fastserde/gettingstarted.md
+++ b/avro-fastserde/gettingstarted.md
@@ -71,8 +71,4 @@ FastSpecificDatumWriter<T> fastSpecificDatumWriter = new FastSpecificDatumWriter
 fastSpecificDatumWriter.write(data, binaryEncoder);
 
 
-## Troubleshooting
-
-
-## Related Documentation
 
--- a/avro-fastserde/gettingstarted.md
+++ b/avro-fastserde/gettingstarted.md
@@ -1,0 +1,78 @@
+Getting Started
+===============
+
+## Introduction
+
+This lib is a fork of the open-source project: https://github.com/RTBHOUSE/avro-fastserde.
+You could get more information regarding the original thinking of this lib by this comprehensive
+blog: https://techblog.rtbhouse.com/2017/04/18/fast-avro/.
+
+In the high-level, this is an alternative of the standard Avro serialization/de-serialization
+framework, but with better performance by doing runtime code generation.
+According to the prod experiment, the de-serlialization time could be reduced by 90% for
+some complicated schema, and the serialization time is also improved substantially.
+
+If your application is encountering performance issue with Avro serialization/de-serialization,
+you might want to give it a try.
+
+When bringing this lib to LinkedIn, we have done the following improvements so far:
+1. Made the lib work with both Avro-1.4 and Avro-1.7;
+2. Modified the lib to return "Utf8" instead of "String" to align with
+the standard Avro lib;
+3. Improved the performance of the original lib by caching the generated
+de-serlializer in "FastSpecificDatumReader" and "FastGenericDatumReader";
+4. Improved the unit tests to always run the tests against both this fast lib
+and the standard Avro lib;
+5. Improved the de-serialization generator to support object reuse;
+6. Introduced some GC optimization for float vectors;
+
+We will continue to improve this lib in the future since we are seeing this lib could
+be widely used.
+
+
+## Requirements
+
+## Installation/Setting Up
+
+Gradle dependency:
+com.linkedin.avroutil:avro-codegen:0.1.1
+
+## Using the Product or Service
+
+Since all the APIs in this lib is compatible with the generic Avro lib, so it is very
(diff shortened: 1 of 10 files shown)


gaojieliu:
@FelixGV @radai-rosenblatt 
Please review.

avro-fastserde/gettingstarted.md
```diff
@@ -0,0 +1,78 @@
...
+fastSpecificDatumReader.read(reuse, binaryDecoder);
+
+FastSpecificDatumWriter<T> fastSpecificDatumWriter = new FastSpecificDatumWriter<>(schema);
+fastSpecificDatumWriter.write(data, binaryEncoder);
+
+
+## Troubleshooting
+
+
+## Related Documentation
```
FelixGV: I think we should remove this instead of leaving those sections empty.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerializerGeneratorBase.java
```diff
@@ -60,22 +61,25 @@ protected static int nextUniqueInt() {
   public abstract FastSerializer<T> generateSerializer();
 
   @SuppressWarnings("unchecked")
-  protected Class<FastSerializer<T>> compileClass(final String className) throws IOException, ClassNotFoundException {
+  protected Class<FastSerializer<T>> compileClass(final String className, Set<String> knownUsedFullyQualifiedClassNameSet)
```
FelixGV: Looks like this class and `FastDeserializerGeneratorBase` have a lot in common. We should consider refactoring the code so they share a common base class or share utility functions...

I don't think it's critical to do this as part of this commit, but we should at least leave a TODO comment.

I leave it up to you whether you want to refactor right away or just leave the TODO (unless you disagree and think the two classes ought to stay separate?)

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
```diff
@@ -159,7 +159,7 @@ public Float get(int i) {
    *
    * TODO: verify if it's enough to simply store the data as primitives in order to benefit from the GC optimization.
    */
-  private boolean addPrimitive(float o) {
+  public boolean addPrimitive(float o) {
```
FelixGV: Why did we need this change? It doesn't seem related to this commit, right?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/SchemaAssistant.java
```diff
@@ -25,20 +25,32 @@
...
+   */
+  private final Set<String> fullyQualifiedClassNameSet = new HashSet<>();
 
   public SchemaAssistant(JCodeModel codeModel, boolean useGenericTypes) {
     this.codeModel = codeModel;
     this.useGenericTypes = useGenericTypes;
     this.exceptionsFromStringable = new HashSet<>();
   }
 
+  protected Set<String> getUsedFullyQualifiedClassNameSet() {
```
FelixGV: Doesn't seem really "schema related". Should we move this functionality to `Utils` instead?

avro-fastserde/gettingstarted.md
gaojieliu: Sure, I can do that.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerializerGeneratorBase.java
gaojieliu: You are right, and this situation has been there from the beginning, and we haven't optimized the code structure yet.
I could leave a TODO.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
gaojieliu: This is required by LinkedIn internal application.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/SchemaAssistant.java
gaojieliu: This set is used to capture all the used classes during code generation (SchemaAssistant#classFromSchema is the function to extract all the used classes), and I think this is the right place to get this piece of info.

PR #17, opened 2020-01-30
Fast Serializer support for Avro-1.4

gaojieliu:
This code change adds support for fast serializer of Avro-1.4.
Now, fast serializer/de-serializer are available for
Avro 1.4/1.7/1.8.

This special logic for Avro-1.4 in Fast Serializer generation
is for enum/fixed since those two types in Avro-1.4 are implementing
differently from other versions, and the corresponding Schema couldn't
be extracted from the generic implementations (GenericData.EnumSymbol
and GenericData.Fixed).

@FelixGV @radai-rosenblatt


avro-fastserde/src/main/java/com/linkedin/avro/fastserde/SchemaAssistant.java
```diff
@@ -67,9 +71,10 @@ public static boolean isComplexType(Schema schema) {
   public static boolean isNamedType(Schema schema) {
     switch (schema.getType()) {
       case RECORD:
+        return true;
       case ENUM:
       case FIXED:
-        return true;
+        return Utils.isAvro14() ? false : true;
```
radai-rosenblatt: avro 1.4 spec says fixed _are_ named types - https://avro.apache.org/docs/1.4.0/spec.html#Names
what does this condition affect ?

radai-rosenblatt: if this is to avoid generating a getSchema() call (which 1.4 fixed classes dont have), i'd at least put a comment.

gaojieliu: You are right, I added a comment to mention the purpose.

PR #27, opened 2020-02-27
add convenience method for creating a BinaryDecoder on top of a byte[]

radai-rosenblatt:


helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
```diff
@@ -122,11 +123,20 @@ public static BinaryDecoder newBinaryDecoder(InputStream in, boolean buffered, B
     return ADAPTER.newBinaryDecoder(in, buffered, reuse);
   }
 
+  /**
+   * convenience method for getting a {@link BinaryDecoder} for a given byte[]
+   * @param in byte array with data
+   * @return a {@link BinaryDecoder} for decoding the given array
+   */
+  public static BinaryDecoder newBinaryDecoder(byte[] in) {
+    return newBinaryDecoder(new ByteArrayInputStream(in), false, null);
```
gaojieliu: Shall we use `true` as the default since Buffered BinaryDecoder could provide better performance?
What about the `newBinaryEncoder` function? Are you planning to use Buffered as the default?

radai-rosenblatt: i had it set to true to begin with, but the byte[] is already in memory, so no point?

PR #29, opened 2020-03-16
Change log level to debug in FastDatumReader/Writer to improve perf

volauvent:
Previous log level in FastGenericDatumReader/Writer was "info" which affected performance greatly when trying to reuse cached fastDeserializer/Serializer in FastSerdeCache.

By removing log, we can take better advantage of cached fastDeserializer/Serializer.

According one of the JMH benchmark results:
Fast-deserialization latency can be improved from **1.1 ms** to **0.5ms** by this change.

@gaojieliu @FelixGV


gaojieliu:
LGTM.

PR #31, opened 2020-03-21
Change to use IdentityHashMap to cache schemaId for Avro 1.4

volauvent:
In Avro 1.4, Schema#hashCode is not performance efficient. It has been improved
by caching computed hashcode in the later avro version.

Change to use IdentityHashMap for Avro 1.4 helps to improve the schemaId cache
lookup performance. Here's JMH benchmark of deserialization latency of a example
schema using different approaches:

>  - vanilla avro 1.4:                                                                                     0.26ms
>  - avro-fastserde with ConcurrentHashMap as schemaId cache:   0.22ms
>  - avro-fastserde with IdentityHashMap as schemaId cache:          0.14ms

However, it requires the client to resuse the schame instance. Or, it will eventually
lead to "Too many schema objects" issue. I think it's ok for now since a typical
client generally reuse the schema instance and that's one of the best practices as well.
We can build a LRU Cache with size limit in the future.

@gaojieliu @FelixGV @radai-rosenblatt

### DIFF
--- a/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/Utils.java
+++ b/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/Utils.java
@@ -11,10 +11,12 @@ import java.security.CodeSource;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
+import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.stream.Collectors;
 import org.apache.avro.Schema;
 
@@ -25,6 +27,16 @@ public class Utils {
 
   // Cache the mapping between Schema and the corresponding fingerprint
   private static final Map<Schema, Long> SCHEMA_IDS_CACHE = new ConcurrentHashMap<>();
+
+  // Schema#hashCode is not performance efficient in Avro 1.4. It has been improved
+  // by caching hashcode in the later version. Change to use IdentityHashMap for
+  // Avro 1.4 helps to improve the cache lookup performance. However, it requires
+  // the client to resuse the schame instance. Or, it will ventually lead to
+  // "Too many schema objects" issue
+  // TODO - LRU Cache with size limit
+  private static final Map<Schema, Long> SCHEMA_IDS_CACHE_AVRO_14 = new IdentityHashMap<>();
+
+  private static ReentrantReadWriteLock reentrantReadWriteLock = new ReentrantReadWriteLock();
 
   private Utils() {
   }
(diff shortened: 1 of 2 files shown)


avro-fastserde/src/main/java/com/linkedin/avro/fastserde/Utils.java
```diff
@@ -26,6 +28,16 @@
   // Cache the mapping between Schema and the corresponding fingerprint
   private static final Map<Schema, Long> SCHEMA_IDS_CACHE = new ConcurrentHashMap<>();
 
+  // Schema#hashCode is not performance efficient in Avro 1.4. It has been improved
+  // by caching hashcode in the later version. Change to use IdentityHashMap for
+  // Avro 1.4 helps to improve the cache lookup performance. However, it requires
+  // the client to resuse the schame instance. Or, it will ventually lead to
+  // "Too many schema objects" issue
+  // TODO - LRU Cache with size limit
+  private static final Map<Schema, Long> SCHEMA_IDS_CACHE_AVRO_14 = new IdentityHashMap<>();
```
radai-rosenblatt: as you stated, this has a memory issue. you could use WeakIdentityHashMap for this. while its not a jdk class there are implementations in various places you could port over - https://docs.jboss.org/jbossas/javadoc/4.0.4/common/org/jboss/util/collection/WeakIdentityHashMap.html https://www.gnu.org/software/kawa/api/gnu/kawa/util/WeakIdentityHashMap.html even avro 1.8 has one - https://avro.apache.org/docs/1.8.0/api/java/org/apache/avro/util/WeakIdentityHashMap.html (although they warn its non standard)

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/Utils.java
```diff
@@ -26,6 +28,16 @@
...
+
+  private static ReentrantReadWriteLock reentrantReadWriteLock = new ReentrantReadWriteLock();
```
radai-rosenblatt: nit - make this final

radai-rosenblatt:
good catch with avro 1.4 schemas not caching hashCode(). this is definitely something we can fix at code-gen time - opened #32

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/Utils.java
radai-rosenblatt: yet another potential implementation - https://github.com/gfx/Java-WeakIdentityHashMap

volauvent: Good idea. Done.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/Utils.java
volauvent: Fixed

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/Utils.java
```diff
@@ -26,6 +28,15 @@
   // Cache the mapping between Schema and the corresponding fingerprint
   private static final Map<Schema, Long> SCHEMA_IDS_CACHE = new ConcurrentHashMap<>();
 
+  // Schema#hashCode is not performance efficient in Avro 1.4. It has been improved
+  // by caching hashcode in the later version. Change to use WeakIdentityHashMap for
+  // Avro 1.4 helps to improve the cache lookup performance. However, it requires
+  // the client to resuse the schame instance.
```
radai-rosenblatt: nit - comments about need for LRU no longer true?

gaojieliu:
I have some small concern here.
I prefer not to change the default behavior in fast-avro since it could work no matter whether the schema instance passed is immutable or not. With this change, if the same schema instance gets changed, fast-avro will return wrong serializer/de-serializer..
Avro-1.7 and above could dynamically detect whether the schema instance is changed or not to decide whether to re-generate the hashcode on-demand, so it will always be correct. 

I recommend to create a different set of functions as utilities to clearly mentioned the pre-requisites to use those new functions, which will leverage IdentityHashMap for fast lookup.
We could do it in Fast DatumReader/DatumWriter layer.

Let us know your thoughts.

volauvent:
@gaojieliu  Yeah, it is a concern. I didn't know if the user changes schema instance on-the-fly with broken changes. I thought the only change was usually version field...

I think a new util interface just for avro-1.4 de/serializer cache using IdentityHashMap is a heavy solution and it also brings maintenance work in the future. 

We also can wait for "caching hashCode() in avro-1.4" as @radai-rosenblatt mentioned above. Then, the  Schema.hashcode will not be a performance issue for HashMap anymore.

I will decline this PR then. Thanks!

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/Utils.java
volauvent: We still need this improvement in the future. It' not only related to this PR. The de/serializer cache will also lead to memory issue if user have created a great number of schemas, such as in projection scenario...

gaojieliu:
@volauvent 
Sure.
For the optimization specifically in avro-1.4, I think it is still useful since inside LinkedIn, there are a lot of customers, who are using avro-1.4, and there is no timeline to let all of them migrate to avro-1.7 or above.
I think it is worth exposing such utility functions, otherwise, each customer needs to develop their own solution if they want to bypass this inefficient Schema#hashCode logic in Avro-1.4.

@radai-rosenblatt 
Since you mentioned that this could be avoided by code generation. For specific record, it is possible, but what about for generic record? With the generic record, the code generation couldn't be involved from my understanding, right?

volauvent:
Thanks for review! I will decline this PR and create another one for this perf improvement.

PR #33, opened 2020-03-23
Make avro-fastserde depend on helper module instead of avro-migration-helper module

huangminchn:
Also renamed 3 Avro19SchemaNormalization files to SchemaNormalization.


avro-fastserde/build.gradle
```diff
@@ -31,7 +31,8 @@ if (USE_AVRO_14) {
 }
 
 dependencies {
-  compile project(":avro-migration-helper")
+  compile project(":helper:helper")
+  compile project(":helper:impls:helper-impl-14")
```
radai-rosenblatt: why is this required?

huangminchn: Need this function `SchemaNormalization#fingerprint64` from ":helper:impls:helper-impl-14".

radai-rosenblatt: i think its better to pull up SchemaNormalization into something like com.linkedin.avroutil1.compatibility.SchemaNormalization in the helper module, and consider that API

huangminchn: Done. Please check the latest commit. Thanks!

radai-rosenblatt: so is this line still there? github shows it is

huangminchn: Good catch! This line is not needed any more. Removed.

PR #34, opened 2020-03-24
remove avro-fastserde's declared dependency on helper from pom

radai-rosenblatt:


avro-fastserde/build.gradle
```diff
@@ -145,3 +151,111 @@ cleanupAndRebuildTestsForAvro18.dependsOn generateAvroClasses18
...
+
+      //remove the dependency on helper, and replace with one on helper-all
+      pom.withXml {
+        groovy.util.Node dependenciesNode = (groovy.util.Node) (asNode().dependencies[0])
+        Collection<groovy.util.Node> dependencyNodes = dependenciesNode.children();
+        for (groovy.util.Node o : dependencyNodes) {
+          if ("$project.group" == o.groupId[0].text()) {
+            dependencyNodes.remove(o)
+            break;
+          }
```
FelixGV: So... as soon as we find one dependency with our group name, we remove it and stop looking further? Is this going to cause weird issues in the future if we refactor the build or otherwise add a new module and then somethings start depending on more than one avro-util dependency? Should we also look for whether the name is `helper` in that if condition?

avro-fastserde/build.gradle
```diff
@@ -145,3 +151,111 @@ cleanupAndRebuildTestsForAvro18.dependsOn generateAvroClasses18
...
+        }
+        def dependencyNode = dependenciesNode.appendNode('dependency')
+        dependencyNode.appendNode('groupId', "$project.group")
+        dependencyNode.appendNode('artifactId', "helper-all")
+        dependencyNode.appendNode('version', "$project.version")
+        dependencyNode.appendNode('scope', "compile")
```
FelixGV: Nitpick: `compile` is deprecated in modern versions of Gradle, in favor of `implementation` and `api`. In most cases, `implementation` is the right scope to use.

build.gradle
```diff
@@ -27,7 +27,7 @@ plugins {
 group = 'com.linkedin.avroutil1'
 
 //does not include "helper" as that module has its own publishing section (because fat jar)
```
FelixGV: Can we leave another comment here to say why fastserde can't be treated via the default process?

avro-fastserde/build.gradle
radai-rosenblatt: this is a maven scope though

PR #36, opened 2020-03-26
Performance improvement of PrimitiveFloatList by lazy deserialization

majisourav99:
Improving the deserialization performance of PrimitiveFloatList. 
1. Using ByteBuffer instead of byte[] . ByeBuffer#getFloat has faster performance than avro readFloat.
2. I am moving the deserialization to array access time only, so that clients can deserialize the records extremely fast. The full deserialization happens only when client tries to access array element. Also cache the deserialized elements into the float array after the first access so later accesses will not see any impact.
Regular fast-avro performance of 1M records of 10kb sized float array:  ~18s
Using the lazy deserialization using ByteBuffer: ~ 2s

### DIFF
--- a/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/CompositeByteBuffer.java
+++ b/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/CompositeByteBuffer.java
@@ -1,0 +1,57 @@
+package com.linkedin.avro.fastserde;
+
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.util.ArrayList;
+import java.util.List;
+
+
+public class CompositeByteBuffer {
+  private int byteBufferCount;
+  private List<ByteBuffer> byteBuffers;
+
+  public CompositeByteBuffer() {
+    byteBuffers = new ArrayList<>(2);
+  }
+
+  public ByteBuffer allocate(int index, int size) {
+    ByteBuffer byteBuffer;
+
+    // Check if we can reuse the old record's byteBuffers, else allocate a new one.
+    if (byteBuffers.size() > index && byteBuffers.get(index).capacity() > size) {
+      byteBuffer = byteBuffers.get(index);
+      byteBuffer.clear();
+    } else {
+      byteBuffer = ByteBuffer.allocate((int)size).order(ByteOrder.LITTLE_ENDIAN);
+    }
+    if (index < byteBuffers.size()) {
+      byteBuffers.set(index, byteBuffer);
+    } else {
+      byteBuffers.add(byteBuffer);
+    }
+    return byteBuffer;
+  }
+
+  public void clear() {
+    for (ByteBuffer byteBuffer : byteBuffers)  {
+      byteBuffer.clear();
+    }
+  }
+
+  public void setByteBufferCount(int count) {
+    byteBufferCount = count;
+  }
+
+  public float readFloat(int i) {
+    return byteBuffers.get(0).getFloat(i*4);
+  }
+  public void setArray(float[] array) {
+    int k = 0;
+    for (int i = 0; i < byteBufferCount; i++) {
+      ByteBuffer byteBuffer = byteBuffers.get(i);
+      for (int j = 0; j < byteBuffer.limit(); j += Float.BYTES) {
+        array[k++] = byteBuffer.getFloat(j);
+      }
+    }
+  }
+}
--- a/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/CompositeByteBuffer.java
+++ b/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/CompositeByteBuffer.java
@@ -42,9 +42,6 @@ public class CompositeByteBuffer {
     byteBufferCount = count;
   }
 
-  public float readFloat(int i) {
-    return byteBuffers.get(0).getFloat(i*4);
-  }
   public void setArray(float[] array) {
     int k = 0;
     for (int i = 0; i < byteBufferCount; i++) {
--- a/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
+++ b/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
@@ -1,9 +1,13 @@
 package com.linkedin.avro.fastserde;
 
 import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
 import java.util.AbstractList;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Iterator;
+import java.util.List;
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericArray;
 import org.apache.avro.generic.GenericData;
@@ -25,21 +29,32 @@ import org.apache.avro.io.Decoder;
  * - It re-implements {@link #compareTo(GenericArray)}, {@link #equals(Object)} and {@link #hashCode()}
  *   in order to leverage the primitive types, rather than causing unintended boxing.
  *
- * TODO: Provide arrays for other primitive types.
+ *   Using ByteBuffer to speed up float-array deserialization: We allocate ByteBuffer to store the raw bytes from
+ *   BinaryDecoder and deserialize them only during array element access. We cache the results into the elements array
+ *   after the first get access of the array so that sub-sequent array access are fast.
+ *
+ *   TODO: Provide arrays for other primitive types.
  */
 public class PrimitiveFloatList extends AbstractList<Float>
     implements GenericArray<Float>, Comparable<GenericArray<Float>> {
   private static final float[] EMPTY = new float[0];
+  private static final int FLOAT_SIZE = Float.BYTES;
   private static final Schema FLOAT_SCHEMA = Schema.create(Schema.Type.FLOAT);
   private static final Schema SCHEMA = Schema.createArray(FLOAT_SCHEMA);
-
   private int size;
+  private int byteBufferCount;
   private float[] elements = EMPTY;
+  private List<ByteBuffer> byteBuffers;
+  private boolean isCached = false;
 
   public PrimitiveFloatList(int capacity) {
     if (capacity != 0) {
       elements = new float[capacity];
     }
+  }
+
+  public PrimitiveFloatList() {
+    byteBuffers = new ArrayList<>(2);
   }
 
   public PrimitiveFloatList(Collection<Float> c) {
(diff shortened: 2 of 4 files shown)


avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
```diff
@@ -25,23 +31,32 @@
...
+  private boolean isCached = false;
 
   public PrimitiveFloatList(int capacity) {
     if (capacity != 0) {
       elements = new float[capacity];
     }
   }
 
+  public PrimitiveFloatList() {
+    byteBuffers = new ArrayList<>();
```
radai-rosenblatt: nit - put a better bound on size? (default is 10)

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
```diff
@@ -60,44 +75,72 @@ public PrimitiveFloatList(Collection<Float> c) {
...
+        array.byteBuffers.add(byteBuffer);
+        totalLenth += length;
+        length = binaryDecoder.arrayNext();
+      } while (length > 0);
+
+      setupElements(old, array, (int)totalLenth);
       return array;
     } else {
-      return newPrimitiveFloatArray(old, 0);
+      return new PrimitiveFloatList(0);
```
radai-rosenblatt: is this immutable? if so can this be made a constant? (like COllections.emptyList())

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
```diff
@@ -60,44 +75,72 @@ public PrimitiveFloatList(Collection<Float> c) {
...
+    if (length > 0) {
+      PrimitiveFloatList array = (PrimitiveFloatList) newPrimitiveFloatArray(old);
+
       do {
-        for (long i = 0; i < l; i++) {
-          array.addPrimitive(in.readFloat());
+        // Allocate ByeBuffer of size array length * Float.BYTES to hold the array values
+        ByteBuffer byteBuffer = ByteBuffer.allocate((int)length*4).order(ByteOrder.LITTLE_ENDIAN);
+        ByteBufferBinaryDecoder binaryDecoder;
+        if (in instanceof ByteBufferBinaryDecoder) {
```
radai-rosenblatt: this inctanceOf can be executed once, and not in a loop

majisourav99:
> Instead of doing `* 4` here and `<< 2` later, can we instantiate a `int byteLength = length << 2` and then keep reusing it?

Sure. will do that

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
```diff
@@ -212,14 +258,28 @@ public Float remove(int i) {
...
+    cacheFromByteBuffer();
     Float result = elements[i];
     --size;
     System.arraycopy(elements, i + 1, elements, i, (size - i));
     elements[size] = 0;
     return result;
   }
 
+  private void cacheFromByteBuffer() {
+    if (!isCached) {
```
FelixGV: Looks like this function is not thread-safe, right? Can we add a synchronized block inside this if, which re-checks the `if (!isCached)` condition once inside the synchronized?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
```diff
@@ -60,44 +76,63 @@ public PrimitiveFloatList(Collection<Float> c) {
...
+     *         false otherwise
+     */
   public static boolean isFloatArray(Schema expected) {
     return expected != null && Schema.Type.ARRAY.equals(expected.getType()) && FLOAT_SCHEMA.equals(
         expected.getElementType());
   }
 
-  private static Object newPrimitiveFloatArray(Object old, int size) {
+  private static Object newPrimitiveFloatArray(Object old) {
     if (old instanceof PrimitiveFloatList) {
```
FelixGV: Should `isCached` be flipped back to false when re-using an old instance? Also, this should probably happen inside a synchronized block? Let's think about how to reduce the critical section to the smallest possible number of lines...

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
```diff
@@ -25,23 +31,33 @@
...
  */
 public class PrimitiveFloatList extends AbstractList<Float>
     implements GenericArray<Float>, Comparable<GenericArray<Float>> {
   private static final float[] EMPTY = new float[0];
+  private static final int FLOAT_SIZE = Float.BYTES;
   private static final Schema FLOAT_SCHEMA = Schema.create(Schema.Type.FLOAT);
   private static final Schema SCHEMA = Schema.createArray(FLOAT_SCHEMA);
-
   private int size;
   private float[] elements = EMPTY;
```
FelixGV: Is there any advantage to caching? How about making deserialization lazy every single time the array is accessed? Can we run a benchmark where we access the array more than once, and try to assess if the caching brings a benefit? It would be nice if we could avoid instantiating this array.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
```diff
@@ -25,23 +31,33 @@
...
  *   in order to leverage the primitive types, rather than causing unintended boxing.
  *
- * TODO: Provide arrays for other primitive types.
+ *   Using ByteBuffer to speed up float-array deserialization: We allocate ByteBuffer to store the raw bytes from
+ *   BinaryDecoder and deserialize them only during array element access. We cache the results into the elements array
+ *   after the first get access of the array so that sub-sequent array access are fast.
+ *
+ *   TODO: Provide arrays for other primitive types.
  */
 public class PrimitiveFloatList extends AbstractList<Float>
```
FelixGV: Instead of changing this class, could we change `PrimitiveFloatList` to an interface, make the existing class one of the implementations, and then create additional implementations? I think this would be cleaner, and would allow us to benchmark the various implementations against one another in a nicer way. It would also make it very clear for our users which methods are intended to be public as part of the interface.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
```diff
@@ -60,44 +75,72 @@ public PrimitiveFloatList(Collection<Float> c) {
...
+
+      setupElements(old, array, (int)totalLenth);
       return array;
     } else {
-      return newPrimitiveFloatArray(old, 0);
+      return new PrimitiveFloatList(0);
     }
   }
 
+  private static void setupElements(Object old, PrimitiveFloatList list, int size) {
```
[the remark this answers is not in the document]
FelixGV: Also, it's not clear to me why we're passing both `old` and `list` here, which, IIUC, are the same instance...?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
```diff
@@ -60,44 +76,63 @@ public PrimitiveFloatList(Collection<Float> c) {
...
+      if (size <= oldFloatList.getCapacity()) {
+        // reuse the float array directly
+        oldFloatList.clear();
+      } else {
+        oldFloatList.resizeAndClear(size);
+      }
+      return;
+    }
+    list.elements = new float[size];
+    list.size = size;
```
FelixGV: Looks like this is the only place where the `size` member variable is set, however, this code will not execute if the early termination above occurs, in which case, `size` would not be set... I think we should make sure to always set the `size` in the beginning, prior to deserializing any of the items.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
```diff
@@ -60,44 +76,63 @@ public PrimitiveFloatList(Collection<Float> c) {
...
+
       do {
-        for (long i = 0; i < l; i++) {
-          array.addPrimitive(in.readFloat());
-        }
-        l = in.arrayNext();
-      } while (l > 0);
+        // Allocate ByeBuffer of size array length * Float.BYTES to hold the array values
+        ByteBuffer byteBuffer = ByteBuffer.allocate((int)length * FLOAT_SIZE).order(ByteOrder.LITTLE_ENDIAN);
+        in.readFixed(byteBuffer.array(), 0, (int)(length * FLOAT_SIZE));
```
FelixGV: This will allocate a new `byte[]` inside the `ByteBuffer` and copy the content from the `Decoder` into it, right? Can we manage to get hold of the original `byte[]` that was passed to the `Decoder` instead, and have that one back the `ByteBuffer`, without any copy?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
```diff
@@ -25,23 +31,33 @@
...
+ *   Using ByteBuffer to speed up float-array deserialization: We allocate ByteBuffer to store the raw bytes from
+ *   BinaryDecoder and deserialize them only during array element access. We cache the results into the elements array
+ *   after the first get access of the array so that sub-sequent array access are fast.
+ *
+ *   TODO: Provide arrays for other primitive types.
  */
 public class PrimitiveFloatList extends AbstractList<Float>
     implements GenericArray<Float>, Comparable<GenericArray<Float>> {
   private static final float[] EMPTY = new float[0];
+  private static final int FLOAT_SIZE = Float.BYTES;
```
FelixGV: Doesn't matter from a performance point of view, but is this necessary? Using `Float.BYTES` directly in our code seems pretty clean and readable to me.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
```diff
@@ -212,14 +266,28 @@ public Float remove(int i) {
...
     elements[size] = 0;
     return result;
   }
 
+  private void cacheFromByteBuffer() {
+    if (!isCached) {
+      for (ByteBuffer byteBuffer : byteBuffers) {
+        int length = byteBuffer.limit() >> 2;
+        for (int i = 0; i < length; i++) {
+          elements[i] = byteBuffer.getFloat(i << 2);
```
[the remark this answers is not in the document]
majisourav99: > Are these two bit shifts necessary? Why not do the following?
> 
> ```
>         for (int i = byteBuffer.position(); i < byteBuffer.limit(); i += Float.BYTES) {
>           elements[i] = byteBuffer.getFloat(i);
>         }
> ```
> 
> Note that we should not assume that the buffer's position is zero...

That code will not work, will make holes in the array. I could use two variables, also ByteBuffer does start from 0 in this case.

int k = 0;
for (int i = byteBuffer.position(); i < byteBuffer.limit(); i += Float.BYTES) {
  elements[k++] = byteBuffer.getFloat(i);
}

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
```diff
@@ -60,44 +75,73 @@ public PrimitiveFloatList(Collection<Float> c) {
...
       do {
-        for (long i = 0; i < l; i++) {
-          array.addPrimitive(in.readFloat());
+        ByteBuffer byteBuffer;
+        // Allocate ByeBuffer of size array length * Float.BYTES to hold the array values
+        if (array.byteBuffers.size() > cnt && (array.byteBuffers.get(cnt).limit() > (length * FLOAT_SIZE))) {
+          byteBuffer = array.byteBuffers.get(cnt);
+          byteBuffer.clear();
+        } else {
+          byteBuffer = ByteBuffer.allocate((int)length * FLOAT_SIZE).order(ByteOrder.LITTLE_ENDIAN);
```
FelixGV: I don't think it'll make any difference performance-wise, but I think the code would look cleaner if we instantiated `int bytesLength = length << 2` and plugged it into the various places of the code...

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
```diff
@@ -60,44 +75,73 @@ public PrimitiveFloatList(Collection<Float> c) {
...
+      array.byteBufferCount = cnt;
+      setupElements(array, (int)totalLength);
       return array;
     } else {
-      return newPrimitiveFloatArray(old, 0);
+      return new PrimitiveFloatList(0);
+    }
+  }
+
+  private static void setupElements(PrimitiveFloatList list, int total_size) {
```
FelixGV: Can we say avoid the underscore syntax in favor of camelCase, which is more typical in Java?

majisourav99:
> I think this is pretty much as good as it gets with the current approach. I left some minor comments. It might be worthwhile to explore some other variations, including an option to return immutable objects, as that could relax some of the constraints you're facing here by needing to uphold the complete `List<Float>` contract...

I agree code would be much cleaner if we didnt support add/remove, butI am keeping this for now as its implements more generic interface for users to use.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
gaojieliu: That is a good question.
@majisourav99 
Could you do a quick benchmark?
Without caching, the complexity here will be reduced greatly.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
```diff
@@ -60,44 +76,74 @@ public PrimitiveFloatList(Collection<Float> c) {
...
-        for (long i = 0; i < l; i++) {
-          array.addPrimitive(in.readFloat());
+        ByteBuffer byteBuffer;
+        long byteSize = length * FLOAT_SIZE;
+        // Check if we can reuse the old record's byteBuffers, else allocate a new one.
+        if (array.byteBuffers.size() > cnt && array.byteBuffers.get(cnt).limit() > byteSize) {
+          byteBuffer = array.byteBuffers.get(cnt);
+          byteBuffer.clear();
+        } else {
+          byteBuffer = ByteBuffer.allocate((int)byteSize).order(ByteOrder.LITTLE_ENDIAN);
```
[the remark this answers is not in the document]
majisourav99: Acutally Direct buffer does not expose the array() method, which would not allow me to read the bytes from the Decoder.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
```diff
@@ -60,44 +76,74 @@ public PrimitiveFloatList(Collection<Float> c) {
...
+      PrimitiveFloatList array = (PrimitiveFloatList) newPrimitiveFloatArray(old);
+      int cnt = 0;
+
       do {
-        for (long i = 0; i < l; i++) {
-          array.addPrimitive(in.readFloat());
+        ByteBuffer byteBuffer;
+        long byteSize = length * FLOAT_SIZE;
+        // Check if we can reuse the old record's byteBuffers, else allocate a new one.
+        if (array.byteBuffers.size() > cnt && array.byteBuffers.get(cnt).limit() > byteSize) {
```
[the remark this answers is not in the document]
majisourav99: Yes, we should check for capacity(). I will fix it

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
majisourav99: I tried reading the array twice and caching vs non-caching gives around ~20% better performance. But caching is here mostly to support the full arrayList interface of add/remove functionality.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
majisourav99: Done

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
majisourav99: wish I could do that. BinaryDecoder does not give us access for that. Probably better approach would to fork BinaryDecoder which would give access to the raw bytes.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
majisourav99: Done

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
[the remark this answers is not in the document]
majisourav99: cleaned up the code.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
majisourav99: done

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
majisourav99: Probably not do it as for reuse case we could reuse it.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/CompositeByteBuffer.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+  }
+
+  public void setByteBufferCount(int count) {
+    byteBufferCount = count;
+  }
+
+  public float readFloat(int i) {
+    return byteBuffers.get(0).getFloat(i*4);
+  }
```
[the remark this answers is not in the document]
majisourav99: Thanks for catching this, it was for testing caching vs non-caching performance. I will remove it.

PR #35, opened 2020-03-24
JMH benchmark suite for avro-fastserde

volauvent:
Use JMH suite benchmark performance of vanilla avro and fast-serde serialization and deserialization. It ueses AvroRandomDataGenerator to generate avro data according to specified Schema.

To run this benchmark: `./gradlew :avro-fastserde:jmh -PUSE_AVRO_14/17/18`

Then it will benchmark the average serdes latency and print GC details of vanilla Avro and fast-serdes.

You also can test your own AVRO schema by replacing the schema in
`avro-util/avro-fastserde/src/test/avro/benchmarkSchema.avsc`

JMH benchmark is very helpful for -
1. Benchmark the serdes latency reduction of customer schema brought by applying avro-fastserde
2. Investigate in the performance improvement and bottleneck of different schema pattern and avro version

@gaojieliu @FelixGV

PR #37, opened 2020-03-26
Fast Datum Reader/Writer Util with WeakIdentityHashMap

pravk03:
FastDatumReaderWriterUtil is a wrapper utility to create Fast Datum Reader and Writer objects . The Utility internally takes care caching the created objects in order to avoid creation of objects from the same Schema / Schema pair . 

This will help address the inefficiency with Schema#hashCode computation in Avro 1.4 

Pre-requisite to use this util:
 1. Schema object will not be changed on-the-FLY, or the changes do NOT require new DatumReader/Writer
  2. Your application will always use the same Schema object for the same schema

Latency improvement by caching : 
Comparing with creating new Datum reader / write object for every record 
Deser : ~40% 
Ser  : ~20%

@volauvent @FelixGV @radai-rosenblatt @gaojieliu

### DIFF
--- a/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDatumReaderWriterUtil.java
+++ b/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDatumReaderWriterUtil.java
@@ -48,10 +48,10 @@ public class FastDatumReaderWriterUtil {
   private static final ReentrantReadWriteLock reentrantReadWriteLock = new ReentrantReadWriteLock();
 
   //TODO :  LRU cache
-  private static final Map<SchemaPair, FastGenericDatumReader> fastGenericDatumReaderCache = new ConcurrentHashMap<>();
+  private static final Map<SchemaPair, FastGenericDatumReader> fastGenericDatumReaderCache = new FastAvroConcurrentHashMap<>();
   private static final Map<Schema, FastGenericDatumWriter> fastGenericDatumWriterCache = new WeakIdentityHashMap();
 
-  private static final Map<SchemaPair, FastSpecificDatumReader> fastSpecificDatumReaderCache = new ConcurrentHashMap<>();
+  private static final Map<SchemaPair, FastSpecificDatumReader> fastSpecificDatumReaderCache = new FastAvroConcurrentHashMap<>();
   private static final Map<Schema, FastSpecificDatumWriter> fastSpecificDatumWriterCache = new WeakIdentityHashMap();
 
   private FastDatumReaderWriterUtil() {
@@ -78,8 +78,8 @@ public class FastDatumReaderWriterUtil {
     }
     // update cache and write lock
     if (fastDatumWriter == null) {
+      reentrantReadWriteLock.writeLock().lock();
       try {
-        reentrantReadWriteLock.writeLock().lock();
         fastDatumWriter = new FastGenericDatumWriter<>(writerSchema);
         fastGenericDatumWriterCache.put(writerSchema, fastDatumWriter);
       } finally {
@@ -111,8 +111,8 @@ public class FastDatumReaderWriterUtil {
     }
     // update cache and write lock
     if (fastDatumWriter == null) {
+      reentrantReadWriteLock.writeLock().lock();
       try {
-        reentrantReadWriteLock.writeLock().lock();
         fastDatumWriter = new FastSpecificDatumWriter<>(writerSchema);
         fastSpecificDatumWriterCache.put(writerSchema, fastDatumWriter);
       } finally {
--- a/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDatumReaderWriterUtil.java
+++ b/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDatumReaderWriterUtil.java
@@ -48,31 +48,31 @@ public class FastDatumReaderWriterUtil {
   private static final ReentrantReadWriteLock reentrantReadWriteLock = new ReentrantReadWriteLock();
 
   //TODO :  LRU cache
-  private static final Map<SchemaPair, FastGenericDatumReader> fastGenericDatumReaderCache = new FastAvroConcurrentHashMap<>();
-  private static final Map<Schema, FastGenericDatumWriter> fastGenericDatumWriterCache = new WeakIdentityHashMap();
+  private static final Map<SchemaPair, FastGenericDatumReader<?>> fastGenericDatumReaderCache = new FastAvroConcurrentHashMap<>();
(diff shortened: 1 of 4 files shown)


avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDatumReaderWriterUtil.java
```diff
@@ -0,0 +1,124 @@
...
+  private FastDatumReaderWriterUtil() {
+  }
+
+  public static FastGenericDatumReader getFastGenericDatumReader(Schema schema) {
+    return getFastGenericDatumReader(schema, schema);
+  }
+
+  public static FastGenericDatumReader getFastGenericDatumReader(Schema writerSchema, Schema readerSchema) {
+    SchemaPair schemaPair = new SchemaPair(writerSchema, readerSchema);
+    return fastGenericDatumReaderCache.computeIfAbsent(schemaPair, key -> new FastGenericDatumReader<>(writerSchema, readerSchema));
```
gaojieliu: ConcurrentHashMap is not very efficient in this way, check this JDK ticket:
https://bugs.openjdk.java.net/browse/JDK-8161372

Let us create a wrapper of ConcurrentHashMap, here is an example being used in our project:
```
public V computeIfAbsent(K key, Function<? super K, ? extends V> mappingFunction) {
    V value = get(key);
    if (value != null) {
      return value;
    }
    return super.computeIfAbsent(key, mappingFunction);
  }
```

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDatumReaderWriterUtil.java
```diff
@@ -0,0 +1,124 @@
...
+  public static FastSpecificDatumReader getFastSpecificDatumReader(Schema schema) {
+    return getFastSpecificDatumReader(schema, schema);
+  }
+
+  public static FastSpecificDatumReader getFastSpecificDatumReader(Schema writerSchema, Schema readerSchema) {
+    SchemaPair schemaPair = new SchemaPair(writerSchema, readerSchema);
+    return fastSpecificDatumReaderCache.computeIfAbsent(schemaPair, key -> new FastSpecificDatumReader<>(writerSchema, readerSchema));
+  }
+
+  public static FastSpecificDatumWriter getFastSpecificDatumWriter(Schema writerSchema) {
```
gaojieliu: This function is almost same as `getFastGenericDatumWriter`, so we can create a utility function to wrap up the common logic, and the common function could take a `supplier` callback to create the actual DatumWriter.

FelixGV:
> Hi @pravk03, can you also attach the serdes latency reduction metrics by caching FastGenericDatumReader/Writer in your JMH benchmark test?

Better yet, can we contribute the benchmark to the project (: ?

volauvent:
The PR for benchmark is already ready to review :) Check out the PR#35 https://github.com/linkedin/avro-util/pull/35

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDatumReaderWriterUtil.java
```diff
@@ -0,0 +1,124 @@
...
+    reentrantReadWriteLock.readLock().lock();
+    try {
+      fastDatumWriter = fastGenericDatumWriterCache.get(writerSchema);
+    } finally {
+      reentrantReadWriteLock.readLock().unlock();
+    }
+    // update cache and write lock
+    if (fastDatumWriter == null) {
+      try {
+        reentrantReadWriteLock.writeLock().lock();
```
radai-rosenblatt: nit - should be outside the try

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDatumReaderWriterUtil.java
pravk03: Done . Added a Wrapper class and using that instead of ConcurrentHashMap .

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDatumReaderWriterUtil.java
pravk03: done

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDatumReaderWriterUtil.java
pravk03: Yes . The functions are identical . We use different cache for generic and specific types though .

volauvent:
Thanks for the benchmark result!
> Comparing with creating new Datum Reader/Write object for every record, latency reduction -
> Deser : ~40%
> Ser : ~20%
LGTM. Please squash commits before merge :)

pravk03:
> Thanks for the benchmark result!
> 
> > Comparing with creating new Datum Reader/Write object for every record, latency reduction of test schema with Avro 1.4 -
> > Deser : ~40%, from 2.4ms to 1.3ms;
> > Ser : ~20%, from 3.5ms to 2.7ms;
> 
> LGTM. Please squash commits before merge :)

gaojieliu:
@pravk03 Has this change got merged yet?

volauvent:
@gaojieliu  Thanks for review. This PR has not been merged yet.
cc @pravk03

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDatumReaderWriterUtil.java
gaojieliu: That is right, but it is still possible to pass the suppliers to define how to retrieve from the cache/put into the cache.
The goal here is trying to minimize  duplicate  codes.
The common function could be:
DatumWriter<T> getFastDatumWriter(Schema schema, Supplier<DatumWriter> readFromCache, Consumer<DatumWriter> putIntoCache);
With this way, you need to do some static casting to make it work.

BTW, why did you remove the generic type in the signature of all those functions?

gaojieliu:
Got it, thanks.
I left one more follow-up comment, and please help evaluate it.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDatumReaderWriterUtil.java
pravk03: Thanks for pointing this out @gaojieliu  . Have used generic types .  
Totally agree with the points about having a common getFastDatumWriter()  . Will take that up as an improvement .

PR #41, opened 2020-04-06
Primt float array enhancement

majisourav99:
Setup composite byte buffer even for general non-avro List<Float> interface implementaion of  PrimitiveFloatList usage.
 eg
PrimitiveFloatList list = new PrimitiveFloatList(10);

 It will not use the ByteBuffer code, but cacheFromByteBuffer tries to read from the ByteBuffer array, which would fail if it is not initialized.


majisourav99:
> Instead of doing `* 4` here and `<< 2` later, can we instantiate a `int byteLength = length << 2` and then keep reusing it?

majisourav99:
> Instead of doing `* 4` here and `<< 2` later, can we instantiate a `int byteLength = length << 2` and then keep reusing it?

PR #42, opened 2020-04-07
Initialize ByteBuffer for PrimitiveFloatList List<Float> interface usage

majisourav:
Setup composite byte buffer even for general non-avro List<Float> interface implementation of PrimitiveFloatList usage.
eg
PrimitiveFloatList list = new PrimitiveFloatList(10);

The above use-case will not use the ByteBuffer data buffer, but cacheFromByteBuffer tries to read from the ByteBuffer to fill in float array, which would fail if it is not initialized.

### DIFF
--- a/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
+++ b/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
@@ -48,10 +48,8 @@ public class PrimitiveFloatList extends AbstractList<Float>
     if (capacity != 0) {
       elements = new float[capacity];
     }
-  }
-
-  public PrimitiveFloatList() {
-    byteBuffer = new CompositeByteBuffer();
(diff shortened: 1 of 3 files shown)


avro-fastserde/src/main/java/com/linkedin/avro/fastserde/PrimitiveFloatList.java
```diff
@@ -59,6 +56,7 @@ public PrimitiveFloatList(Collection<Float> c) {
       elements = new float[c.size()];
       addAll(c);
     }
+    byteBuffer = new CompositeByteBuffer();
```
gaojieliu, line 59: Since we know we won't use it, can we initialize it with an empty list in `CompositeByteBuffer` for this scenario?
`Collections.emptyList`.

majisourav: > Since we know we won't use it, can we initialize it with an empty list in `CompositeByteBuffer` for this scenario?
> `Collections.emptyList`.

Sure.

PR #44, opened 2020-04-13
Fixed a bug in processUnion of FastSerializer

gaojieliu:
Previously, the code is checking schema type to decide the right index,
and this change was introduced to accommodate Avro-1.4, but this is
not appropriate since Union could contain multiple record types, which
will cause the writen index will always be the index of the first record.
The fix is to check the full name of the schema instead of type name.

@volauvent @FelixGV

### DIFF
--- a/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/SchemaAssistant.java
+++ b/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/SchemaAssistant.java
@@ -68,6 +68,15 @@ public class SchemaAssistant {
     }
   }
 
+  public static String getSchemaFullName(Schema schema) {
+    Schema.Type type = schema.getType();
+    boolean isNamedType = type.equals(Schema.Type.ENUM) || type.equals(Schema.Type.FIXED) || type.equals(Schema.Type.RECORD);
+    /**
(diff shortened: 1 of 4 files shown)


avro-fastserde/src/main/java/com/linkedin/avro/fastserde/SchemaAssistant.java
```diff
@@ -68,6 +68,15 @@ public static boolean isComplexType(Schema schema) {
     }
   }
 
+  public static String getSchemaFullName(Schema schema) {
+    Schema.Type type = schema.getType();
+    boolean isNamedType = type.equals(Schema.Type.ENUM) || type.equals(Schema.Type.FIXED) || type.equals(Schema.Type.RECORD);
+    /**
+     * Avro-1.4 doesn't support {@link Schema#getFullName()} if the Schema is not a NamedSchema.
```
radai-rosenblatt, line 75: EnumSchema extends NamedSchema even in 1.4

volauvent:
LGTM, thx for the fixup.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/SchemaAssistant.java
gaojieliu: Yes, that is why it is allowed in the code path of Avro-1.4.

PR #45, opened 2020-04-13
PrimitiveGenericRecord interface

FelixGV:


avro-fastserde/src/main/java/com/linkedin/avro/primitive/PrimitiveGenericRecord.java
```diff
@@ -0,0 +1,103 @@
...
+   * }
+   *
+   * This allows optional fields to be handled as primitives and thus avoid the boxing
+   * overhead.
+   *
+   * @param key the field name
+   * @return true if the field has a value defined, or false if it is null
+   * @throws AvroRuntimeException
+   */
+  boolean isDefined(String key) throws AvroRuntimeException;
```
radai-rosenblatt, line 54: nit - "isDefined" sounds (to me?) more like if the schema contains a field. maybe hasValue()? of nonNull() ?

avro-fastserde/src/main/java/com/linkedin/avro/primitive/PrimitiveGenericRecord.java
```diff
@@ -0,0 +1,103 @@
+package com.linkedin.avro.primitive;
```
radai-rosenblatt, line 1: nit - if we're gonna go extending avro ... "com.linkedin.avro.extensions" ?

volauvent:
Nice! I think it definitely can improve deser performance further like PrimitiveFloatList.
@radai-rosenblatt  Here is the performance improvement brought by PrimitiveFloatList benchmarked by a Float Array with 800 elements -

> - vanilla avro 1.4
> _Deser latency (ns) :                   24603.5
> Memory alloc (KB/exe) :             15.9_
> - avro-fastserde with **PrimitiveFloatList**
> _Deser latency (ns) :                   3926.4
> Memory alloc (KB/exe) :            3.2_
> - avro-fastserde with **GenericArray\<Float\>**
> _Deser latency (ns) :                  6904.2
> Memory alloc (KB/exe) :            15.8_

avro-fastserde/src/main/java/com/linkedin/avro/primitive/PrimitiveGenericRecord.java
FelixGV: Yeah, I wasn't sure how to name this one. Also thought of `isSet(name)`... what's your favorite wording?

avro-fastserde/src/main/java/com/linkedin/avro/primitive/PrimitiveGenericRecord.java
FelixGV: Sure, or maybe `com.linkedin.avro.api`?

radai-rosenblatt: works for me

avro-fastserde/src/main/java/com/linkedin/avro/primitive/PrimitiveGenericRecord.java
radai-rosenblatt: what if its set to null? :-)

gaojieliu:
@volauvent 
For primitive list, the big benefit is totally reasonable.
But for the primitive type for each field, we may need to do another round of benchmark to understand the benefit accurately, and it should be related to how many primitive types the users are planning to store inside the record.
Can we try different number of primitive fields and evaluate the benefit?

@FelixGV 
Also do we recommend the usage to everyone or just the ones who are heavily using primitive fields in the record types? Since I could see another coding overhead while using this new type, I would like to understand the proper use case.

volauvent:
@gaojieliu since we only have primitiveFloatList now, we can vary float array size and compare the performance diff between Primitive float and default box one. The above metric is for 800 FloatArray, are you saying reduce the array size to a small number to understand the real performance improvement brought by primitive float type, such as to 5?

PR #48, opened 2020-05-13
Fix a bug in object reuse in CompositeByteBuffer

majisourav:
The buffer size should be less than equal to the new size for object reuse to utilize the existing buffer in CompositeByteBuffer. The bug did not check for equality.


avro-fastserde/src/test/java/com/linkedin/avro/fastserde/FastDeserializerGeneratorForReuseTest.java
```diff
@@ -207,11 +207,13 @@ public void testFastGenericDeserializerPrimitFloatList() throws Exception {
     // new record array length shorter than reuse record.
     arrayList.clear();
     arrayList.add((float)10);
+    arrayList.add((float)20);
+    arrayList.add((float)30);
```
FelixGV, line 211: This change seems to contradict the comment...? Shall we change the comment or the code?

majisourav: Sure. I will update the comment

PR #49, opened 2020-05-17
Improve fast deserialization by avoiding field schema retrieval cost

volauvent:
Improved fast deserialization speed by avoiding current field schema retrieval cost.
Changed to use fields' schema directly instead of registering and then retrieving
them from HashMap.

JMH benchmark results of fast-deserialization time
1. Enum array with 200 elements
```
        |   Avro 1.4(ns)   |  Avro 1.8(ns)
Before  |   7452           |   13101 
After   |   5374           |   7858 
```

2. Record array with 200 elements
```
        |   Avro 1.4(ns)   |   Avro 1.8(ns)
Before  |   23068          |   24519
After   |   17854          |   18549
```

@gaojieliu @FelixGV @radai-rosenblatt

### DIFF
--- a/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
+++ b/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
@@ -9,7 +9,6 @@ import com.sun.codemodel.JConditional;
 import com.sun.codemodel.JDoLoop;
 import com.sun.codemodel.JExpr;
 import com.sun.codemodel.JExpression;
-import com.sun.codemodel.JFieldVar;
 import com.sun.codemodel.JForLoop;
 import com.sun.codemodel.JInvocation;
 import com.sun.codemodel.JMethod;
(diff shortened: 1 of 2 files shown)


volauvent:
Here are deserializers of EnumArray schema:

Generated deserializer before this PR:
https://gist.github.com/volauvent/76e0f69131bf2ce5b9eeba6a72319d85

Generated deserializer after this PR:
https://gist.github.com/volauvent/13ca4ed55b48e761c8b8e474f0a5da4b

FelixGV:
Nice results! The generated code LGTM. The code generator is a bit puzzling. I've seen nothing wrong but I'm not confident that I've looked at everything in-depth enough... in any case, if the tests pass and Gaojie thinks it's good, then I guess it's good enough.

BTW, this is off-topic, but do we have a grasp of why 1.8 would perform slower than 1.4?

gaojieliu:
Sorry, I forgot to follow up, and the generated de-serializer looks good!

volauvent:
> BTW, this is off-topic, but do we have a grasp of why 1.8 would perform slower than 1.4?

The reason why previous fast-avro 1.8 EnumArray deserialization performs much slower than 1.4 is that Avro 1.8 uses _Schema_ and _Symbol_ to construct `EnumSymbol` while avro 1.4 only uses _Symbol_.

It leads to fast-deserializer 1.8 suffers 2X cost of retrieving schema from HashMap as bellow
```
// deserializer in avro 1.8
enumArray2 .add(new org.apache.avro.generic.GenericData.EnumSymbol(
readerSchemaMap.get(4483722390578694240L), 
readerSchemaMap.get(4483722390578694240L).getEnumSymbols().get((decoder.readEnum()))));

// deserializer in avro 1.4
enumArray2 .add(new org.apache.avro.generic.GenericData.EnumSymbol(
readerSchemaMap.get(4483722390578694240L).getEnumSymbols().get((decoder.readEnum()))));
```
@FelixGV

PR #52, opened 2020-05-26
add newBinaryDecoder which re-configures a BinaryDecoder in AvroCompatibilityHelper

xinxinzhenhuai:
Venice has a local optimization requires re-initialize a BinaryDecoder, but the method signature is different in avro 1.4 and versions after 1.4. In order to be compatible with all avro versions, we need to add this method in AvroCompatibilityHelper. Created several classes of AvroXXBinaryDecoderConfigurer since init/configure method in BinaryDecoder is package-private.

### DIFF
--- a/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvroAdapter.java
+++ b/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvroAdapter.java
@@ -37,6 +37,9 @@ public interface AvroAdapter {
 
   BinaryDecoder newBinaryDecoder(ObjectInput in);
 
+  BinaryDecoder newBinaryDecoder(byte[] bytes, int offset,
+      int length, BinaryDecoder reuse);
+
   JsonEncoder newJsonEncoder(Schema schema, OutputStream out, boolean pretty) throws IOException;
 
   JsonDecoder newJsonDecoder(Schema schema, InputStream in) throws IOException;
--- a/helper/impls/helper-impl-14/src/main/java/com/linkedin/avroutil1/compatibility/avro14/Avro14Adapter.java
+++ b/helper/impls/helper-impl-14/src/main/java/com/linkedin/avroutil1/compatibility/avro14/Avro14Adapter.java
@@ -36,6 +36,7 @@ import org.apache.avro.Avro14SchemaAccessUtil;
 import org.apache.avro.AvroRuntimeException;
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericData;
+import org.apache.avro.io.Avro14BinaryDecoderAccessUtil;
 import org.apache.avro.io.BinaryDecoder;
 import org.apache.avro.io.BinaryEncoder;
 import org.apache.avro.io.DecoderFactory;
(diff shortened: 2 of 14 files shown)


helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvroAdapter.java
```diff
@@ -37,6 +37,9 @@
 
   BinaryDecoder newBinaryDecoder(ObjectInput in);
 
+  BinaryDecoder configureBinaryDecoder(byte[] bytes, int offset,
+      int length, BinaryDecoder reuse);
```
FelixGV, line 41: I recommend aligning the naming with that of the other functions: `newBinaryEncoder`, which also support a `reuse` parameter. It should be valid to pass null in the `reuse` param, and therefore, this function doesn't necessarily only configure an existing decoder, but it is also allowed to create a new one, in some cases.

helper/impls/helper-impl-14/src/main/java/com/linkedin/avroutil1/compatibility/avro14/Avro14Adapter.java
```diff
@@ -106,6 +109,11 @@ public BinaryDecoder newBinaryDecoder(ObjectInput in) {
     return newBinaryDecoder(new ObjectInputToInputStreamAdapter(in), false, null);
   }
 
+  @Override
+  public BinaryDecoder configureBinaryDecoder(byte[] bytes, int offset, int length, BinaryDecoder reuse) {
+    return configurer.configureBinaryDecoder(bytes, offset, length, reuse);
```
FelixGV, line 114: Why do we need a Configurer? Why not do it inline in this class like in `newBinaryEncoder(OutputStream out, boolean buffered, BinaryEncoder reuse)`?

BTW, I think we need to have the same checks as in the above function to deal with null `reuse` instances and also to check that the type is one we expect...

xinxinzhenhuai: because the method `newBinaryEncoder ` is using is public, but the method we are using is package-private and is not accessible in linkedin package.

FelixGV: Ooooh I see. Thanks. Please add a JavaDoc comment on the Configurer classes to point that out.

helper/impls/helper-impl-14/src/main/java/org/apache/avro/io/Avro14BinaryDecoderConfigurer.java
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright 2020 LinkedIn Corp.
+ * Licensed under the BSD 2-Clause License (the "License").
+ * See License in the project root for license information.
+ */
+
+package org.apache.avro.io;
+
+public class Avro14BinaryDecoderConfigurer {
```
radai-rosenblatt, line 9: the convention is to call these classes *AccessUtil - so Avro14BinaryDecoderAccessUtil?
also - can be static

PR #53, opened 2020-05-28
Always return PrimitiveFloatList, even when cold

FelixGV:
The PrimitiveFloatList is an API which users should expect to rely on,
so it is wrong to degrade to the more constrained List<Float> API while
Fast-Avro is still cold. This commit introduces several changes to make
the extended API reliably present whenever using Fast-Avro, regardless
of being cold or warm.

- Changed PrimitiveFloatList to an interface, in a new package called:
  com.linkedin.avro.api; since the package name migration makes this an
  incompatible change, it would be desirable for the next release to
  not increment only the patch version. Having a proper package name
  for API extension should make things cleaner in the future as we add
  other optimized APIs (e.g. PR #45).
- Renamed the old class to ByteBufferBackedPrimitiveFloatList, and made
  it implement the new interface.
- Added new several new classes to ensure that the PrimitiveFloatList
  is always returned even when Fast-Avro falls back to vanilla Avro:
  - ColdPrimitiveFloatList which is a naive implementation that simply
    implements the new API by delegating to the regular Avro functions.
    This does not provide any GC benefits, but at least maintains the
    API.
  - ColdGenericDatumReader and ColdSpecificDatumReader which extend the
    GenericDatumReader and SpecificDatumReader classes, respectively,
    from vanilla Avro.
  - ColdDatumReaderMixIn which provides a utility function to minimize
    repeated code between the two DatumReader functions.
- Significantly refactored the FastGenericDeserializerGeneratorTest so
  that it tests three permutations: vanilla, cold fast and warm fast.
  As part of doing this, several test short-comings were discovered and
  fixed. In particular, the decodeRecordSlow function had some flipped
  parameters which led to test failures on vanilla Avro, and those test
  failures were hidden by the fact that some tests ignored the provided
  permutation param and systematically tested only Fast-Avro.

### DIFF
--- a/avro-fastserde/src/main/java/com/linkedin/avro/api/PrimitiveFloatList.java
+++ b/avro-fastserde/src/main/java/com/linkedin/avro/api/PrimitiveFloatList.java
@@ -1,0 +1,20 @@
+package com.linkedin.avro.api;
+
+import java.util.List;
+
+/**
+ * A {@link List} implementation with additional functions to prevent boxing.
+ */
+public interface PrimitiveFloatList extends List<Float> {
+  /**
+   * @param index index of the element to return
+   * @return the element at the specified position in this list
+   */
+  float getPrimitive(int index);
+
+  /**
+   * @param e element whose presence in this collection is to be ensured
+   * @return <tt>true</tt> if this collection changed as a result of the call
+   */
+  boolean addPrimitive(float e);
+}
--- a/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/ByteBufferBackedPrimitiveFloatList.java
+++ b/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/ByteBufferBackedPrimitiveFloatList.java
@@ -1,5 +1,6 @@
 package com.linkedin.avro.fastserde;
 
+import com.linkedin.avro.api.PrimitiveFloatList;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.AbstractList;
@@ -33,8 +34,8 @@ import org.apache.avro.io.Decoder;
  *
  *   TODO: Provide arrays for other primitive types.
  */
-public class PrimitiveFloatList extends AbstractList<Float>
-    implements GenericArray<Float>, Comparable<GenericArray<Float>> {
+public class ByteBufferBackedPrimitiveFloatList extends AbstractList<Float>
+    implements GenericArray<Float>, Comparable<GenericArray<Float>>, PrimitiveFloatList {
   private static final float[] EMPTY = new float[0];
   private static final int FLOAT_SIZE = Float.BYTES;
   private static final Schema FLOAT_SCHEMA = Schema.create(Schema.Type.FLOAT);
@@ -44,7 +45,7 @@ public class PrimitiveFloatList extends AbstractList<Float>
   private boolean isCached = false;
   private CompositeByteBuffer byteBuffer;
 
-  public PrimitiveFloatList(int capacity) {
+  public ByteBufferBackedPrimitiveFloatList(int capacity) {
     if (capacity != 0) {
       elements = new float[capacity];
     }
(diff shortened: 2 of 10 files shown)


avro-fastserde/src/main/java/com/linkedin/avro/api/PrimitiveFloatList.java
```diff
@@ -0,0 +1,11 @@
+package com.linkedin.avro.api;
+
+import java.util.List;
+
+/**
+ * A {@link List} implementation with additional functions to prevent boxing.
+ */
+public interface PrimitiveFloatList extends List<Float> {
+  float getPrimitive(int index);
+  boolean addPrimitive(float o);
```
radai-rosenblatt, line 10: worth documenting the boolean return value of add (its non trivial, at least for me)

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/ByteBufferBackedPrimitiveFloatList.java
```diff
@@ -90,11 +91,11 @@ public static Object readPrimitiveFloatArray(Object old, Decoder in) throws IOEx
       setupElements(array, (int)totalLength);
       return array;
     } else {
-      return new PrimitiveFloatList(0);
+      return new ByteBufferBackedPrimitiveFloatList(0);
```
radai-rosenblatt, line 94: will the result be mutated? maybe this can be constified?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/coldstart/ColdPrimitiveFloatList.java
```diff
@@ -0,0 +1,22 @@
+package com.linkedin.avro.fastserde.coldstart;
+
+import com.linkedin.avro.api.PrimitiveFloatList;
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericData;
+
+public class ColdPrimitiveFloatList extends GenericData.Array<Float> implements PrimitiveFloatList {
```
radai-rosenblatt, line 7: javadoc - "implements PrimitiveFloatList by delegating to the normal (expensive) calls" or something

avro-fastserde/src/main/java/org/apache/avro/generic/ColdDatumReaderMixIn.java
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * An interface with default implementation in order to defeat the lack of multiple inheritance.
+ */
+public interface ColdDatumReaderMixIn {
+  default Object newArray(Object old, int size, Schema schema, NewArrayFunction fallBackFunction) {
+    switch (schema.getElementType().getType()) {
+      case FLOAT:
+        if (null == old || !(old instanceof ColdPrimitiveFloatList)) {
+          old = new ColdPrimitiveFloatList(size);
```
radai-rosenblatt, line 15: nit - cam simply return here

volauvent:
LGTM. Curious to know what's the exact performance improvement to cold-mode. Worth a JMH benchmark :-)

avro-fastserde/src/main/java/com/linkedin/avro/api/PrimitiveFloatList.java
FelixGV: Sure, I've copied the JavaDoc from `List::get` and from `Collection::add`, respectively.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/ByteBufferBackedPrimitiveFloatList.java
FelixGV: I think we do need to assume the result can be mutated. I'm open to introducing a deserialization mode where the returned objects are read-only though, but that would be orthogonal to this change...

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/coldstart/ColdPrimitiveFloatList.java
FelixGV: Done.

avro-fastserde/src/main/java/org/apache/avro/generic/ColdDatumReaderMixIn.java
FelixGV: Sure. I also spotted a bug in this function... fixed it.

FelixGV:
Actually, there is no performance improvement whatsoever (but there could be some improvements in the future with this method...). The goal of this PR is purely to ensure that the extended API is always provided, even in cold mode.

Issue #50, opened 2020-05-18
Regression in EnumArray fast-avro serialization under Avro 1.4


volauvent:
Fast-avro serialization is slower than regular avro for EnumArray schema under Avro 1.4. Check out below benchmark results and schema pattern. It was faster before version 0.1.9
The serialization under Avro 1.8 is fine. 

**Avro 1.4 benchmark resut:**
```
Benchmark                                                                           Mode  Cnt     Score       Error   Units
FastAvroSerdesBenchmark.testAvroSerialization                                       avgt    5  7431.495 ±   906.343   ns/op
FastAvroSerdesBenchmark.testAvroSerialization:·gc.alloc.rate                        avgt    5   413.867 ±    48.562  MB/sec
FastAvroSerdesBenchmark.testAvroSerialization:·gc.alloc.rate.norm                   avgt    5  3384.000 ±     0.001    B/op
FastAvroSerdesBenchmark.testAvroSerialization:·gc.churn.PS_Eden_Space               avgt    5   389.459 ±   865.511  MB/sec
FastAvroSerdesBenchmark.testAvroSerialization:·gc.churn.PS_Eden_Space.norm          avgt    5  3143.086 ±  6991.629    B/op
FastAvroSerdesBenchmark.testAvroSerialization:·gc.churn.PS_Survivor_Space           avgt    5     0.044 ±     0.315  MB/sec
FastAvroSerdesBenchmark.testAvroSerialization:·gc.churn.PS_Survivor_Space.norm      avgt    5     0.359 ±     2.556    B/op
FastAvroSerdesBenchmark.testAvroSerialization:·gc.count                             avgt    5     4.000              counts
FastAvroSerdesBenchmark.testAvroSerialization:·gc.time                              avgt    5    15.000                  ms
FastAvroSerdesBenchmark.testFastAvroSerialization                                   avgt    5  9664.812 ±  1175.680   ns/op
FastAvroSerdesBenchmark.testFastAvroSerialization:·gc.alloc.rate                    avgt    5   315.128 ±    36.635  MB/sec
FastAvroSerdesBenchmark.testFastAvroSerialization:·gc.alloc.rate.norm               avgt    5  3352.000 ±     0.001    B/op
FastAvroSerdesBenchmark.testFastAvroSerialization:·gc.churn.PS_Eden_Space           avgt    5   259.297 ±   933.239  MB/sec
[...]

PR #54, opened 2020-05-28
Improved fast serializer code generation logic for Enum type

volauvent:
Improved fast serialization speed by avoiding the cost of storing and retrieving Enum schemas
in a Hashmap in Avro 1.4. We have seen ~44% improvement of serialization latency for Enum array.

This PR helps to resolve issue #50 

JMH benchmark results of fast serialization time of an 200 elements EnumArray under Avro 1.4

**Before** 
```
FastAvroSerdesBenchmark.testFastAvroSerialization                                   avgt   10  12054.915 ± 1852.177   ns/op
FastAvroSerdesBenchmark.testFastAvroSerialization:·gc.alloc.rate                    avgt   10    255.063 ±   42.062  MB/sec
FastAvroSerdesBenchmark.testFastAvroSerialization:·gc.alloc.rate.norm               avgt   10   3352.001 ±    0.001    B/op
FastAvroSerdesBenchmark.testFastAvroSerialization:·gc.churn.PS_Eden_Space           avgt   10    272.156 ±  362.710  MB/sec
FastAvroSerdesBenchmark.testFastAvroSerialization:·gc.churn.PS_Eden_Space.norm      avgt   10   3489.396 ± 4814.784    B/op
FastAvroSerdesBenchmark.testFastAvroSerialization:·gc.churn.PS_Survivor_Space       avgt   10      0.103 ±    0.477  MB/sec
FastAvroSerdesBenchmark.testFastAvroSerialization:·gc.churn.PS_Survivor_Space.norm  avgt   10      1.505 ±    6.959    B/op
FastAvroSerdesBenchmark.testFastAvroSerialization:·gc.count                         avgt   10      6.000             counts
FastAvroSerdesBenchmark.testFastAvroSerialization:·gc.time                          avgt   10     46.000                 ms
```

**After**
```
FastAvroSerdesBenchmark.testFastAvroSerialization                                   avgt   10  6709.067 ±  225.951   ns/op
FastAvroSerdesBenchmark.testFastAvroSerialization:·gc.alloc.rate                    avgt   10   453.950 ±   14.888  MB/sec
FastAvroSerdesBenchmark.testFastAvroSerialization:·gc.alloc.rate.norm               avgt   10  3352.000 ±    0.001    B/op
FastAvroSerdesBenchmark.testFastAvroSerialization:·gc.churn.PS_Eden_Space           avgt   10   452.412 ±  247.752  MB/sec
FastAvroSerdesBenchmark.testFastAvroSerialization:·gc.churn.PS_Eden_Space.norm      avgt   10  3337.223 ± 1836.797    B/op
FastAvroSerdesBenchmark.testFastAvroSerialization:·gc.churn.PS_Survivor_Space       avgt   10     0.104 ±    0.480  MB/sec
FastAvroSerdesBenchmark.testFastAvroSerialization:·gc.churn.PS_Survivor_Space.norm  avgt   10     0.802 ±    3.719    B/op
FastAvroSerdesBenchmark.testFastAvroSerialization:·gc.count                         avgt   10     9.000             counts
FastAvroSerdesBenchmark.testFastAvroSerialization:·gc.time                          avgt   10    37.000                 ms
```

### DIFF
--- a/avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/serialization/AVRO_1_4/FastGenericSerializerGeneratorTest_shouldWriteMultipleChoiceUnion_GenericSerializer_4388851848367235159.java
+++ b/avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/serialization/AVRO_1_4/FastGenericSerializerGeneratorTest_shouldWriteMultipleChoiceUnion_GenericSerializer_4388851848367235159.java
@@ -2,10 +2,7 @@
 package com.linkedin.avro.fastserde.generated.serialization.AVRO_1_4;
 
 import java.io.IOException;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 import com.linkedin.avro.fastserde.FastSerializer;
-import org.apache.avro.Schema;
 import org.apache.avro.generic.IndexedRecord;
 import org.apache.avro.io.Encoder;
 import org.apache.avro.util.Utf8;
@@ -14,7 +11,6 @@ public class FastGenericSerializerGeneratorTest_shouldWriteMultipleChoiceUnion_G
     implements FastSerializer<IndexedRecord>
 {
 
-    private Map<Long, Schema> enumSchemaMap = new ConcurrentHashMap<Long, Schema>();
 
     public void serialize(IndexedRecord data, Encoder encoder)
         throws IOException
--- a/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerializerGenerator.java
+++ b/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerializerGenerator.java
@@ -6,7 +6,6 @@ import com.sun.codemodel.JClassAlreadyExistsException;
 import com.sun.codemodel.JConditional;
 import com.sun.codemodel.JExpr;
 import com.sun.codemodel.JExpression;
-import com.sun.codemodel.JFieldVar;
 import com.sun.codemodel.JForEach;
 import com.sun.codemodel.JForLoop;
 import com.sun.codemodel.JMethod;
@@ -16,11 +15,8 @@ import com.sun.codemodel.JVar;
 import java.io.File;
 import java.io.IOException;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
 import org.apache.avro.Schema;
 import org.apache.avro.io.Encoder;
 import org.apache.avro.util.Utf8;
@@ -36,13 +32,9 @@ public class FastSerializerGenerator<T> extends FastSerializerGeneratorBase<T> {
   private final SchemaAssistant schemaAssistant;
 
   /**
-   * Enum schema mapping for Avro-1.4.
+   * Enum schema mapping for Avro-1.4 to record schema id and corresponding schema JVar.
    */
-  private JFieldVar enumSchemaMapField;
-  /**
-   * This field is used to decide whether the corresponding schema is already in {@link #enumSchemaMapField} or not.
-   */
-  private final Set<Long> enumSchemaIdSet = new HashSet<>();
+  private final Map<Long, JVar> enumSchemaVarMap = new HashMap<>();
 
 
   public FastSerializerGenerator(boolean useGenericTypes, Schema schema, File destination, ClassLoader classLoader,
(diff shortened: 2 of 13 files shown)


FelixGV:
Cool result! And how does it compare with vanilla Avro now? Is it equal or better?

Can we get a diff of the generated code? Actually, I'm wondering... should we make the build trigger the code-gen for a set of schemas we care about, and check it under source control (not in the main resource, but perhaps in tests or in some kind of `audit` resource). If we're going to make it part of the workflow to review the generated code, we might as well think of a way to really institutionalize it properly...

volauvent:
Fast-avro 1.4 serialization speed is ~10% faster than vanilla 1.4 now for Enum array. Below is the benchmark result of vanilla avro:
```
FastAvroSerdesBenchmark.testAvroSerialization                                       avgt   10  7432.819 ±  279.983   ns/op
FastAvroSerdesBenchmark.testAvroSerialization:·gc.alloc.rate                        avgt   10   410.783 ±   15.542  MB/sec
FastAvroSerdesBenchmark.testAvroSerialization:·gc.alloc.rate.norm                   avgt   10  3360.000 ±    0.001    B/op
FastAvroSerdesBenchmark.testAvroSerialization:·gc.churn.PS_Eden_Space               avgt   10   402.187 ±  326.169  MB/sec
FastAvroSerdesBenchmark.testAvroSerialization:·gc.churn.PS_Eden_Space.norm          avgt   10  3303.211 ± 2694.281    B/op
FastAvroSerdesBenchmark.testAvroSerialization:·gc.churn.PS_Survivor_Space           avgt   10     0.024 ±    0.094  MB/sec
FastAvroSerdesBenchmark.testAvroSerialization:·gc.churn.PS_Survivor_Space.norm      avgt   10     0.201 ±    0.791    B/op
FastAvroSerdesBenchmark.testAvroSerialization:·gc.count                             avgt   10     8.000             counts
FastAvroSerdesBenchmark.testAvroSerialization:·gc.time                              avgt   10    22.000                 ms
```

Here are fast-serializers of EnumArray schema:
- new fast-serializer before this PR:
https://gist.github.com/volauvent/d1fd5d25788fd8bdd33ba5bcf0ddfd27

- old fast-serializer after this PR:
https://gist.github.com/volauvent/78a117c5ffe0f76ef8ace15f632157f3

Like the idea of source control of code-gen to review related changes more carefully and properly.

FelixGV:
Thanks for the gist... the change looks good. It looks quite weird to have a map lookup/set in this code. I wonder why it was done like this in the first place, and if that's a hint that there may be edge cases where this is important (I can't think of any, though).

Also, you'll notice that there is yet another map lookup hidden in the `getEnumOrdinal` function. This is kind of silly. If the `EnumSymbol` class kept an ordinal field within it, the lookup would be unnecessary. This is non-trivial to change because it would require extending or replacing some of the core Avro classes within `GenericData`, which we haven't done so far, but I had other reasons for wanting to do that in PR #45 so maybe we'll go there at some point... Anyhow, this is clearly out of scope from your current change.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerializerGenerator.java
```diff
@@ -326,21 +304,21 @@ private void processEnum(Schema enumSchema, JExpression enumValueExpression, JBl
...
               getVariableName(enumSchema.getName() + "EnumSchema"),
-              enumSchemaMapField.invoke("get").arg(JExpr.lit(enumSchemaFingerprint)));
-          JConditional schemaCheckCond = body._if(JExpr._null().eq(enumSchemaVar));
-          JBlock thenBody = schemaCheckCond._then();
-          thenBody.assign(enumSchemaVar, codeModel.ref(Schema.class).staticInvoke("parse").arg(enumSchema.toString()));
-          thenBody.invoke(enumSchemaMapField, "put").arg(JExpr.lit(enumSchemaFingerprint)).arg(enumSchemaVar);
-
+              codeModel.ref(Schema.class).staticInvoke("parse").arg(enumSchema.toString()));
+          enumSchemaVarMap.put(enumSchemaFingerprint, enumSchemaVar);
           valueToWrite = JExpr.invoke(enumSchemaVar, "getEnumOrdinal").arg(enumValueCasted.invoke("toString"));
```
FelixGV, line 322: I find this code to be a bit spaghetti... could we instead follow a pattern of 

```
JVar enumSchemaVar = enumSchemaVarMap.computeIfAbsent(enumSchemaFingerprint, () -> 
  serializerClass.field(
              JMod.PRIVATE | JMod.FINAL, 
              Schema.class,
              getVariableName(enumSchema.getName() + "EnumSchema"),
              codeModel.ref(Schema.class).staticInvoke("parse").arg(enumSchema.toString()))
);
valueToWrite = JExpr.invoke(enumSchemaVar, "getEnumOrdinal").arg(enumValueCasted.invoke("toString"));
```

... or something like that? Otherwise, the `JExpr.invoke` part duplicated twice, and since it is meta-code, that makes things extra confusing :D ...

volauvent: Thanks for the suggestion. Fixed.

volauvent:
> I wonder why it was done like this in the first place, and if that's a hint that there may be edge cases where this is important (I can't think of any, though)

Support Avro 1.4 Enum type serialization here was initially introduced by @gaojieliu . Do you have any concern?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerializerGenerator.java
FelixGV: Beautiful! Love the red tint :D !

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/serialization/AVRO_1_4/FastGenericSerializerGeneratorTest_shouldWriteMultipleChoiceUnion_GenericSerializer_4388851848367235159.java
```diff
@@ -14,7 +11,6 @@
     implements FastSerializer<IndexedRecord>
 {
 
-    private Map<Long, Schema> enumSchemaMap = new ConcurrentHashMap<Long, Schema>();
```
FelixGV, line 17: Heh... deleting an unused variable... that sounds like the right call!

PR #55, opened 2020-06-01
New Primitive List APIs

FelixGV:
Introduced 4 new primitive list types similar to PrimitiveFloatList for boolean, 
double, int, long. Also added a "Cold" implementation for each of those, which
is currently returned both in the cold and warm mode. The goal is just to make
sure the primitive API is always available. The performance optimization for
avoiding boxing will come in a future commit.

There is also a change to the meta-code which removes the peeking behavior
for primitive lists. This peeking is useless for primitive types since they are not
mutable, and it results in allocating an Object that will not be used.

In SchemaAssistant:
- Added new isPrimitive() and isCapableOfReuse() functions.
- Modified classFromSchema() to return the new APIs.

Enhanced FastGenericDeserializerGeneratorTest so that it checks that the
getPrimitive() function of the new API works in both cold and warm mode
for all types of primitive lists.


avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/Array_of_BOOLEAN_GenericDeserializer_5988037297456147388_5988037297456147388.java
```diff
@@ -18,36 +18,32 @@ public Array_of_BOOLEAN_GenericDeserializer_5988037297456147388_5988037297456147
...
+                array734 = arrayReuse736;
             } else {
-                array738 = new org.apache.avro.generic.GenericData.Array<Boolean>(((int) chunkLen739), readerSchema);
+                array734 = new ColdPrimitiveBooleanList(((int) chunkLen735));
             }
             do {
-                for (int counter741 = 0; (counter741 <chunkLen739); counter741 ++) {
-                    Object arrayArrayElementReuseVar742 = null;
-                    if ((reuse) instanceof GenericArray) {
-                        arrayArrayElementReuseVar742 = ((GenericArray)(reuse)).peek();
```
FelixGV, line 41: This object allocation caused by the peeking is removed in the new version of the generated code.

radai-rosenblatt: how far do you want to go? if the underlying is a real primitive list Systems.arraycopy() is even faster :-)

FelixGV: I don't think `arraycopy()` applies here... the element peeking happens only so that the element from the old list can be recycled (mutated) with a new value.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -587,97 +569,114 @@ private void processArray(JVar arraySchemaVar, final String name, final Schema a
...
+    ifCodeGen(parentBody, chunkLengthGreaterThanZero, then1 -> {
+      if (finalAction.getShouldRead()) {
+        JExpression conditionForReuse = reuseSupplier.get()
+            /** The reuse param must not be null */
+            .ne(JExpr.direct("null"))
+            /**
+             * And it must be an instance of the right type.
+             * N.B.: Need to use the erasure because instanceof does not support generic types
+             */
+            .cand(reuseSupplier.get()._instanceof(abstractErasedArrayClass));
```
volauvent, line 621: Nitpick: looks like we don't need to validate "null" here since `_instanceof(abstractErasedArrayClass)` also covers null case?

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/FastGenericDeserializerGeneratorTest_shouldReadSubRecordComplexCollectionsField_GenericDeserializer_3807952864887349709_3807952864887349709.java
```diff
@@ -141,13 +137,9 @@ public IndexedRecord deserializeFastGenericDeserializerGeneratorTest_shouldReadS
...
-                        List<IndexedRecord> recordsMapArrayValueReuse0 = null;
-                        if (null instanceof List) {
-                            recordsMapArrayValueReuse0 = ((List) null);
-                        }
-                        if (recordsMapArrayValueReuse0 != (null)) {
-                            recordsMapArrayValueReuse0 .clear();
-                            recordsMapArrayValue0 = recordsMapArrayValueReuse0;
+                        if ((null!= (null))&&(null instanceof List)) {
+                            recordsMapArrayValue0 = ((List) null);
+                            recordsMapArrayValue0 .clear();
```
volauvent, line 142: This part of generated code corresponds to 
```
.ne(JExpr.direct("null"))
.cand(reuseSupplier.get()._instanceof(abstractErasedArrayClass));
```

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
FelixGV: Does it? I didn't know that.

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/Array_of_LONG_GenericDeserializer_2055015354772118358_2055015354772118358.java
```diff
@@ -21,31 +22,23 @@ public Array_of_LONG_GenericDeserializer_2055015354772118358_2055015354772118358
...
-            if ((reuse) instanceof List) {
-                arrayReuse0 = ((List)(reuse));
-            }
-            if (arrayReuse0 != (null)) {
-                arrayReuse0 .clear();
-                array0 = arrayReuse0;
+            if ((reuse) instanceof PrimitiveLongList) {
+                array0 = ((PrimitiveLongList)(reuse));
+                array0 .clear();
             } else {
```
FelixGV, line 31: FYI @volauvent, I adopted your suggestion. Looks very clean now...!

volauvent:
LGTM!

FelixGV:
Thank you!

PR #56, opened 2020-06-03
Checking in samples of fast-avro code gen, to ease future reviews.

FelixGV:
The goal of this change is to have the output of the fast-avro code gen
under version control, so that it can be diffed easily during pull
requests, and so that historical changes are easier to track. Given the
complexity of analyzing  meta-code changes, the hope is that this will
allow a more streamlined and robust workflow.

In order to facilitate this, the main code is changed as such:

- The algorithm for naming the generated classes has been tweaked, so
  that it's more easily recognizable. In particular, collection types
  now specify the type of their element in the class name, e.g.
  - Array_of_record
  - Array_of_LONG
  - Map_of_UNION
- The package name for generated classes is also tweaked, such that it
  includes the Avro version in it, e.g.
  - com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4
  - com/linkedin/avro/fastserde/generated/serialization/AVRO_1_8

Furthermore, the build and tests are more significantly altered:

- A new src directory called codegen is created. It is confgured as a
  source set in the build with proper dependencies, in order to make
  the generated code easy to navigate in the IDE.
- FastGenericDeserializerGeneratorTest and FGSerializerGeneratorTest
  are modified as such:
  - Instead of using a temporary directory for generated serde classes,
    they now write to the new codegen directory.
  - The names of all top level record schemas is changed to include the
    class and function name, in order to facilitate the identification
    of which serde class maps to which test. This is figured out
    automatically via a new utility function in FastSerdeTestsSupport.
- FastGenericDeserializerGeneratorTest now includes 4 new primitive
  array tests which were previously untested: boolean, double, int and
  long.
- The avro-fastserde build now deletes the content of the codegen
  directory before running, to ensure that if test schemas change,
  older serde classes that aren't used anymore will get purged. At the
  end of the build, .class files are also deleted.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,8 @@ build/
 
 #temp artifacts
 avro_tools/
-generated/
+avro-fastserde/src/test/java/com/linkedin/avro/fastserde/generated/
+*.class
 
 #IDEA artifacts
 out/
--- a/avro-fastserde/build.gradle
+++ b/avro-fastserde/build.gradle
@@ -46,6 +46,10 @@ jmh {
   }
 }
 
+sourceSets {
+  codegen
+}
+
 dependencies {
   compile project(":helper:helper")
 
@@ -83,6 +87,11 @@ dependencies {
   avro18 ("org.apache.avro:avro-compiler:1.8.2") {
     exclude group: "org.slf4j"
   }
+
+  codegenCompile ("org.apache.avro:avro-compiler:1.8.2") {
+    exclude group: "org.slf4j"
+  }
+  codegenCompile project(":avro-fastserde")
 }
 
 compileJmhJava {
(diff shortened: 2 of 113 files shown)


volauvent:
LGTM. Yeah, the generated codes should be in the test dir.

I also found it's difficult to read the generated schemas in the `FastGenericDeserializerGeneratorTest` and the `FastGenericSerializerGeneratorTest` during debugging since they are generated manually by calling schema and field creation functions. We can change to define all the test schemas in `.avsc` format under `avro-fastserde/test/avro`. Then in these two tests, use compiled schema classes and `AvroRandomDataGenerator` (current in JMH src root) helper func to generate test records automatically instead of generating them manually. It would be helpful to debug code gen logic. Of course, this improvement is beyond this PR...

volauvent:
Reviewed it again. Looks great to me!

FelixGV:
Thanks @volauvent ! It's merged.

PR #57, opened 2020-06-09
Made fast serde generation deterministic

FelixGV:
This is achieved by altering the unique name generation strategy, such
that it is no longer globally-unique, but rather constrained to being
unique within a given serde class and furthermore within a given name
prefix. This should help remove as much needless jitter as possible in
the codegen output when the meta-code is altered.

Also created a new FastSerdeBase abstract class, extended by both the
FastDeserializerGeneratorBase and FastSerializerGeneratorBase. The
ser and deser meta-code is pretty repetitive, so I'm hoping that we
can move more code to this shared ancestor class later down the line.


volauvent:
LGTM. The codebase is more and more clean now 👍

radai-rosenblatt:
definitely an improvement. 
how hard would it be to add the name of the field being (de)serialized into the variable name as part of the prefix?

FelixGV:
> definitely an improvement.
> how hard would it be to add the name of the field being (de)serialized into the variable name as part of the prefix?

It's definitely doable. Currently there are 24 usages of `getUniqueName()`, so it would take a little while to figure out what is the right bit of info to supplement each one. I think I would prefer moving forward with the the current change for now, but I'm open to iterating further on that, or reviewing someone else's further iteration (: ...

PR #62, opened 2020-06-15
Added support for `avro.java.string` in fast-avro

gaojieliu:
@FelixGV @radai-rosenblatt @volauvent

### DIFF
--- a/avro-fastserde/src/test/java/com/linkedin/avro/fastserde/FastStringableTest.java
+++ b/avro-fastserde/src/test/java/com/linkedin/avro/fastserde/FastStringableTest.java
@@ -14,11 +14,8 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLClassLoader;
-import java.nio.file.Files;
-import java.nio.file.Path;
 import java.util.Collections;
 import org.apache.avro.Schema;
-import org.apache.avro.io.BinaryEncoder;
 import org.apache.avro.io.Decoder;
 import org.apache.avro.io.DecoderFactory;
 import org.apache.avro.io.Encoder;
@@ -46,9 +43,7 @@ public class FastStringableTest {
 
   @BeforeTest(groups = {"deserializationTest", "serializationTest"})
   public void prepare() throws Exception {
-    Path tempPath = Files.createTempDirectory("generated");
-    tempDir = tempPath.toFile();
-
+    tempDir = getCodeGenDirectory();
     classLoader = URLClassLoader.newInstance(new URL[]{tempDir.toURI().toURL()},
         FastDeserializerDefaultsTest.class.getClassLoader());
   }
--- a/avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/StringableRecord_SpecificDeserializer_1822942669987229249_1822942669987229249.java
+++ b/avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/StringableRecord_SpecificDeserializer_1822942669987229249_1822942669987229249.java
@@ -1,0 +1,152 @@
+
+package com.linkedin.avro.fastserde.generated.deserialization.AVRO_1_4;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import com.linkedin.avro.fastserde.FastDeserializer;
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericArray;
+import org.apache.avro.io.Decoder;
+import org.apache.avro.util.Utf8;
+
+public class StringableRecord_SpecificDeserializer_1822942669987229249_1822942669987229249
+    implements FastDeserializer<com.linkedin.avro.fastserde.generated.avro.StringableRecord>
+{
+
+    private final Schema readerSchema;
+
+    public StringableRecord_SpecificDeserializer_1822942669987229249_1822942669987229249(Schema readerSchema) {
+        this.readerSchema = readerSchema;
+    }
+
+    public com.linkedin.avro.fastserde.generated.avro.StringableRecord deserialize(com.linkedin.avro.fastserde.generated.avro.StringableRecord reuse, Decoder decoder)
+        throws IOException
+    {
+        return deserializeStringableRecord0((reuse), (decoder));
+    }
+
+    public com.linkedin.avro.fastserde.generated.avro.StringableRecord deserializeStringableRecord0(Object reuse, Decoder decoder)
+        throws IOException
+    {
+        com.linkedin.avro.fastserde.generated.avro.StringableRecord StringableRecord;
+        if ((reuse)!= null) {
+            StringableRecord = ((com.linkedin.avro.fastserde.generated.avro.StringableRecord)(reuse));
+        } else {
+            StringableRecord = new com.linkedin.avro.fastserde.generated.avro.StringableRecord();
+        }
+        if (StringableRecord.get(0) instanceof Utf8) {
+            StringableRecord.put(0, (decoder).readString(((Utf8) StringableRecord.get(0))));
+        } else {
+            StringableRecord.put(0, (decoder).readString(null));
+        }
+        if (StringableRecord.get(1) instanceof Utf8) {
+            StringableRecord.put(1, (decoder).readString(((Utf8) StringableRecord.get(1))));
+        } else {
+            StringableRecord.put(1, (decoder).readString(null));
+        }
+        if (StringableRecord.get(2) instanceof Utf8) {
+            StringableRecord.put(2, (decoder).readString(((Utf8) StringableRecord.get(2))));
+        } else {
+            StringableRecord.put(2, (decoder).readString(null));
+        }
+        if (StringableRecord.get(3) instanceof Utf8) {
+            StringableRecord.put(3, (decoder).readString(((Utf8) StringableRecord.get(3))));
+        } else {
+            StringableRecord.put(3, (decoder).readString(null));
+        }
+        if (StringableRecord.get(4) instanceof Utf8) {
+            StringableRecord.put(4, (decoder).readString(((Utf8) StringableRecord.get(4))));
+        } else {
+            StringableRecord.put(4, (decoder).readString(null));
+        }
+        List<Utf8> urlArray0 = null;
+        long chunkLen0 = (decoder.readArrayStart());
+        if (chunkLen0 > 0) {
+            if (StringableRecord.get(5) instanceof List) {
+                urlArray0 = ((List) StringableRecord.get(5));
+                urlArray0 .clear();
+            } else {
+                urlArray0 = new ArrayList<Utf8>();
+            }
+            do {
+                for (int counter0 = 0; (counter0 <chunkLen0); counter0 ++) {
+                    Object urlArrayArrayElementReuseVar0 = null;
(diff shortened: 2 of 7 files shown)


FelixGV:
Thanks a lot @gaojieliu!

Could you take a look at the last several commits I made and look at how there is a split between the changes to the meta-code and the changes to the code-gen output? I think it makes the PRs easier to review when we have this split.

I want to propose it as an official development convention for the project but I have not had time to document it yet. If you agree that it makes review easier, please consider adopting the convention and force pushing an update to this PR.

Thanks!

gaojieliu:
@FelixGV 
I adopted your recommendation, and please review it again.

avro-fastserde/src/test/java/com/linkedin/avro/fastserde/FastStringableTest.java
```diff
@@ -137,6 +140,31 @@ public void serializeStringableFields() throws URISyntaxException, MalformedURLE
     }
   }
 
+  @Test(groups = {"deserializationTest"}, dataProvider = "SlowFastDeserializer")
+  public void javaStringPropertyTest(Boolean whetherUseFastDeserializer) throws IOException {
```
volauvent, line 144: Can we also check in the code gen of this test for a better review?

FelixGV: If you wish to include more tests in the code gen under source control, you can use `FastSerdeTestsSupport#getCodeGenDirectory()` and an example of its usage in `FastGenericDeserializerGeneratorTest`.

If you choose to add more stuff to the code gen under source control, then you should also split that into an extra commit (one to add the new code gen, one to modify the meta-code, and another one for the modified code gen) otherwise we can't see the diff if it's all in one commit...

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/Array_of_UNION_GenericDeserializer_585074122056792963_585074122056792963.java
```diff
@@ -74,11 +74,7 @@ public IndexedRecord deserializerecord0(Object reuse, Decoder decoder)
             decoder.readNull();
         }
         if (unionIndex1 == 1) {
-            if (record.get(0) instanceof Utf8) {
-                record.put(0, (decoder).readString(((Utf8) record.get(0))));
-            } else {
-                record.put(0, (decoder).readString(null));
-            }
+            record.put(0, (decoder).readString(((Utf8) record.get(0))));
         }
```
FelixGV, line 78: Hmm... this cast doesn't seem safe. Are you sure that we can remove the `instanceof` check?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeBase.java
```diff
@@ -106,6 +106,7 @@ protected Class compileClass(final String className, Set<String> knownUsedFullyQ
 
     String filePath = destination.getAbsolutePath() + generatedSourcesPath + className + ".java";
     LOGGER.info("Generated source file: " + filePath);
+    System.out.println("Generated source file: " + filePath);
```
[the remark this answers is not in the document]
gaojieliu: Good catch!

avro-fastserde/src/test/java/com/linkedin/avro/fastserde/FastStringableTest.java
gaojieliu: Great idea, I like the way how to store the generated code!

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -897,46 +902,30 @@ private void processBytes(JBlock body, FieldAction action, BiConsumer<JBlock, JE
...
       if (reuseSupplier.get().equals(JExpr._null())) {
-        JExpression readStringInvocation;
-
-        if (schema.getType().equals(Schema.Type.STRING) && !useGenericTypes && SchemaAssistant.isStringable(schema)) {
-          readStringInvocation = JExpr._new(schemaAssistant.classFromSchema(schema))
-              .arg(JExpr.invoke(JExpr.direct(DECODER), "readString").arg(JExpr._null()).
+        argExprForReadString  = JExpr._null();
+      } else {
+        argExprForReadString = JExpr.cast(codeModel.ref(Utf8.class), reuseSupplier.get());
+      }
```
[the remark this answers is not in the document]
gaojieliu: I could move the code to the if section.
And I think it is safe to do `Utf8` cast directly since the passed reuse must pass the schema check before invoking this function, which means if the `stringClass` is decided to be `Utf8`, the non-null reuse must be `Utf8` unless there is a bug.
Let me know if I missed anything  important.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -897,46 +902,30 @@ private void processBytes(JBlock body, FieldAction action, BiConsumer<JBlock, JE
...
+        if (Utils.isAvro14()) {
+          readStringInvocation =
+              JExpr._new(stringClass).arg(JExpr.invoke(JExpr.direct(DECODER), "readString").arg(JExpr._null()).
                   invoke("toString"));
         } else {
-          readStringInvocation = JExpr.invoke(JExpr.direct(DECODER), "readString").arg(JExpr._null());
+          // More GC efficient
+          readStringInvocation =
+              JExpr._new(stringClass).arg(JExpr.invoke(JExpr.direct(DECODER), "readString"));
         }
```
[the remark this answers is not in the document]
gaojieliu: Make sense, will converge towards the way being used in `processMap`, which is more concise.

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/Array_of_UNION_GenericDeserializer_585074122056792963_585074122056792963.java
gaojieliu: Please check my reply above to see whether you  still have this concern or not.

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/FastStringableTest_javaStringPropertyTest_GenericDeserializer_9185875457190242821_9185875457190242821.java
```diff
@@ -0,0 +1,40 @@
...
+package com.linkedin.avro.fastserde.generated.deserialization.AVRO_1_4;
+
+import java.io.IOException;
+import com.linkedin.avro.fastserde.FastDeserializer;
+import org.apache.avro.Schema;
+import org.apache.avro.generic.IndexedRecord;
+import org.apache.avro.io.Decoder;
+import org.apache.avro.util.Utf8;
+
+public class FastStringableTest_javaStringPropertyTest_GenericDeserializer_9185875457190242821_9185875457190242821
```
FelixGV, line 11: If we include those new code-gen outputs as part of the commit after the meta-code is already changed, then we can't see if there was any change to this code path, and if there was, what are they. Are you expecting any changes to this code path? If there are no new kinds of changes exposed by those new code-gen outputs, then I'd rather avoid increasing the footprint of checked in code-gen output, but on the other hand, if there are unique changes here, then it would be good to review them. That is why I mentioned in the other comment thread:

> If you choose to add more stuff to the code gen under source control, then you should also split that into an extra commit (one to add the new code gen, one to modify the meta-code, and another one for the modified code gen) otherwise we can't see the diff if it's all in one commit...

I know this requires some git manipulations, but I think it's important since otherwise changes to the meta-code are very hard to review... LMK if you have concerns.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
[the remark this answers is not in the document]
FelixGV: Hmm... I don't think so. The schema check will ensure that the field is a string, but since the `reuse` object is passed by the application, the field could be populated by any type of `CharSequence`, including either `String` or `Utf8`. Anyway, `instanceof` is quite cheap, from my understanding, and it implicitly includes a null check as well, so I think it'd be better to err on the side of caution and keep the `instanceof`.

We can sync up offline if you want.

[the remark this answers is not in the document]
gaojieliu: Make sense, will add back the `instanceof` check.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -675,6 +675,22 @@ private void processArray(JVar arraySchemaVar, final String name, final Schema a
...
+   * @param stringbleClass
+   * @return
+   */
+  private JExpression readStringableExpression(JClass stringbleClass) {
+    JExpression stringableArgExpr;
+    if (Utils.isAvro14()) {
+      stringableArgExpr = JExpr.direct(DECODER + ".readString(null).toString()");
+    } else {
+      // More GC-efficient
+      stringableArgExpr = JExpr.direct(DECODER + ".readString()");
```
radai-rosenblatt, line 689: can you explain here why avro 1.5+ can be treated differently?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -897,46 +914,22 @@ private void processBytes(JBlock body, FieldAction action, BiConsumer<JBlock, JE
...
-      if (reuseSupplier.get().equals(JExpr._null())) {
-        JExpression readStringInvocation;
-
-        if (schema.getType().equals(Schema.Type.STRING) && !useGenericTypes && SchemaAssistant.isStringable(schema)) {
-          readStringInvocation = JExpr._new(schemaAssistant.classFromSchema(schema))
-              .arg(JExpr.invoke(JExpr.direct(DECODER), "readString").arg(JExpr._null()).
-                  invoke("toString"));
+      JExpression readStringInvocation;
+      JClass stringClass = schemaAssistant.findStringClass(schema);
+      if (stringClass.equals(codeModel.ref(Utf8.class)))  {
```
radai-rosenblatt, line 919: there's an assumption being made here that is the schema says property Record.f should be a string any generated class Record.class "respects" that. this assumption does not generally hold - class Record could have been generated under avro < 1.8 (even if the schema itself has modern annotations)

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
gaojieliu: This function is not  available in avro-1.4 and available in 1.7/1.8 which are versions supported by fast-avro.
I could add some comment for this.

gaojieliu:
@radai-rosenblatt @FelixGV @volauvent 
I am going to close this PR, and create PRs with the right title/commits for this feature.

PR #65, opened 2020-06-22
Disable sharedNameTable for runtime complication

wxn225:
### Problem
The SharedNameTable was introduced to speed up Java complication by using soft references to avoid reallocations. However we have observed severe Memory and GC issue brought by it in fast-avro runtime complication.

The GC issue and heavy memory usage happened when fast-avro needed to process a large number of different schemas. The heap dump of the performance test with 10000 schemas show SharedNameTable consumes ~4GBs memory and it can not be freed from the heap thus causes severe GC issues. Dived into the heap dump further, the SharedNameTable has a static reference to a com.sun.tools.javac.util.List which soft references every single SharedNameTable$NameImpl ever created (so one for each fast serializer that fast-avro has generated and compiled during runtime).

### Solution
SharedNameTable should be disabled for runtime compilation by “-XDuseUnsharedTable” config. The above soft reference memory and GC issue disappeared after disabling SharedNameTable. We also did not observe a noticeable slowness in fast de/serializer code gen.
According to performance test, this memory issue does not exist in Java 11 (tested JDK-11_0_5-zulu and JDK-11_0_5-zing_19_12_100_0_1). We can revert this change in Java 11. However, keeping this config also does not bring any downgrade.

### Reference
https://stackoverflow.com/questions/14617340/memory-leak-when-using-jdk-compiler-at-runtime

### DIFF
--- a/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeBase.java
+++ b/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeBase.java
@@ -112,7 +112,21 @@ public abstract class FastSerdeBase {
     LOGGER.info("For source file: " + filePath + ", and the inferred compile class path: " + compileClassPathForCurrentFile);
     int compileResult;
     try {
-      compileResult = compiler.run(null, null, null, "-cp", compileClassPathForCurrentFile, filePath);
+      /*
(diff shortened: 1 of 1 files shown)


FelixGV:
Thanks a lot for the great investigation and detailed commit message & code comment.

PR #66, opened 2020-06-22
'avro.java.string' support in  fast-avro

gaojieliu:
@radai-rosenblatt @FelixGV @volauvent

### DIFF
--- a/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/SchemaAssistant.java
+++ b/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/SchemaAssistant.java
@@ -5,11 +5,6 @@ import com.linkedin.avro.api.PrimitiveDoubleList;
 import com.linkedin.avro.api.PrimitiveFloatList;
 import com.linkedin.avro.api.PrimitiveIntList;
 import com.linkedin.avro.api.PrimitiveLongList;
-import com.linkedin.avro.fastserde.coldstart.ColdPrimitiveBooleanList;
-import com.linkedin.avro.fastserde.coldstart.ColdPrimitiveDoubleList;
-import com.linkedin.avro.fastserde.coldstart.ColdPrimitiveFloatList;
-import com.linkedin.avro.fastserde.coldstart.ColdPrimitiveIntList;
-import com.linkedin.avro.fastserde.coldstart.ColdPrimitiveLongList;
 import com.linkedin.avro.fastserde.primitive.PrimitiveBooleanArrayList;
 import com.linkedin.avro.fastserde.primitive.PrimitiveDoubleArrayList;
 import com.linkedin.avro.fastserde.primitive.PrimitiveFloatArrayList;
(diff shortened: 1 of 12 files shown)


avro-fastserde/src/main/java/com/linkedin/avro/fastserde/SchemaAssistant.java
```diff
@@ -63,7 +63,7 @@
   public SchemaAssistant(JCodeModel codeModel, boolean useGenericTypes, Class defaultStringClass) {
     this.codeModel = codeModel;
     this.useGenericTypes = useGenericTypes;
-    this.exceptionsFromStringable = new HashSet<>();
+    this.exceptionsFromStringable = new TreeSet<>(Comparator.comparing(Class::getCanonicalName));
```
volauvent, line 66: curious to know why changing to use `TreeSet` here?

avro-fastserde/src/test/java/com/linkedin/avro/fastserde/FastStringableTest.java
```diff
@@ -132,6 +136,30 @@ public void serializeStringableFields() throws URISyntaxException, MalformedURLE
...
+    if (whetherUseFastDeserializer) {
+      afterDecoding = readWithFastAvro(schema, schema, decoder, false);
+    } else {
+      afterDecoding = readWithSlowAvro(schema, schema, decoder, false);
+    }
+
+    if (Utils.isAvro14()){
+      Assert.assertTrue(afterDecoding.get(0) instanceof Utf8, "Utf8 is expected, but got: " + afterDecoding.get(0).getClass());
+    }  else {
+      Assert.assertTrue(afterDecoding.get(0) instanceof String, "String is expected, but got: " + afterDecoding.get(0).getClass());
```
volauvent, line 159: I may miss some context here, so `String` would be the default format after Avro 1.4?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/SchemaAssistant.java
gaojieliu: I should add some comments here, and this is used to make the code gen deterministic in FastDeserializerGenerator.java:
```
for (Class<? extends Exception> classException : schemaAssistant.getExceptionsFromStringable()) {
          JCatchBlock catchBlock = tryBlock._catch(codeModel.ref(classException));
          JVar exceptionVar = catchBlock.param("e");
          catchBlock.body()._throw(JExpr._new(codeModel.ref(AvroRuntimeException.class)).arg(exceptionVar));
        }
```

avro-fastserde/src/test/java/com/linkedin/avro/fastserde/FastStringableTest.java
gaojieliu: Avro-1.4 doesn't recognize the modern hint  defined in the schema: `avro.java.string`, so the de-serialized result will  be Utf8, and those hints were introduced in avro-1.7, and Avro will convert the string field to Java String.
GenericDatumReader.java in avro-1.7:
```
protected Class findStringClass(Schema schema) {
    String name = schema.getProp(GenericData.STRING_PROP);
    if (name == null) return CharSequence.class;

    switch (GenericData.StringType.valueOf(name)) {
      case String:
        return String.class;
      default:
        return CharSequence.class;
    }
  }
```

Issue #67, opened 2020-06-26
slf4j logging API


flowenol:
Hi,

Is there any reason why can't we just use the slf4j API for logging? Currently we are using the apache log4j logger via the dependency for slf4j-log4j12 binding and thus forcing the users to use the log4j binding for slf4j. This is a bit inconvenient because , for example,  in our case this binding is conflicting with log4j over slf4j [bridge](http://www.slf4j.org/legacy.html#log4j-over-slf4j) as we choose to delegate the log4j calls via slf4j to logback in our scenarios. The plain use of slf4j apis does not force users to use any specific flavor for logging.

This unfortunately forces us to exclude the "slf4-log4j12" dependency whenever we are using the avro-fastserde.

FelixGV:
Thanks for the report. You're right, we should use the slf4j APIs rather than the log4j APIs. That's the whole point of slf4j (: ... Feel free to send a PR if you have time. If you can't, we'll get to it ASAP from our side. Cheers!

flowenol:
Hi,

I provided a PR: #69.

I left the log4j APIs in avro-codegen but I assumed that this is a standalone utility so leaving log4j is probably OK.

radai-rosenblatt:
merged, so I'm assuming this can be closed. please re-open if I'm wrong about that

PR #69, opened 2020-06-29
slf4j logging API support in fast-avro (#67)

flowenol:
* slf4j APIs are used now in place of log4j APIs
* slf4j-log4j12 binding dropped from dependencies


radai-rosenblatt:
Thank you!

PR #71, opened 2020-07-02
Fixed a bug when generating specific FastSerializer with Union of Stringable type

gaojieliu:
@FelixGV @radai-rosenblatt @volauvent

### DIFF
--- a/avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_7/StringableRecord_SpecificDeserializer_6174384286732341990_6174384286732341990.java
+++ b/avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_7/StringableRecord_SpecificDeserializer_6174384286732341990_6174384286732341990.java
@@ -1,36 +1,51 @@
 
-package com.linkedin.avro.fastserde.generated.deserialization.AVRO_1_4;
+package com.linkedin.avro.fastserde.generated.deserialization.AVRO_1_7;
 
+import java.io.File;
 import java.io.IOException;
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import com.linkedin.avro.fastserde.FastDeserializer;
+import org.apache.avro.AvroRuntimeException;
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericArray;
 import org.apache.avro.io.Decoder;
-import org.apache.avro.util.Utf8;
 
-public class StringableRecord_SpecificDeserializer_1822942669987229249_1822942669987229249
+public class StringableRecord_SpecificDeserializer_6174384286732341990_6174384286732341990
     implements FastDeserializer<com.linkedin.avro.fastserde.generated.avro.StringableRecord>
 {
 
     private final Schema readerSchema;
 
-    public StringableRecord_SpecificDeserializer_1822942669987229249_1822942669987229249(Schema readerSchema) {
+    public StringableRecord_SpecificDeserializer_6174384286732341990_6174384286732341990(Schema readerSchema) {
         this.readerSchema = readerSchema;
     }
 
     public com.linkedin.avro.fastserde.generated.avro.StringableRecord deserialize(com.linkedin.avro.fastserde.generated.avro.StringableRecord reuse, Decoder decoder)
         throws IOException
     {
-        return deserializeStringableRecord0((reuse), (decoder));
+        try {
+            return deserializeStringableRecord0((reuse), (decoder));
+        } catch (NumberFormatException e) {
+            throw new AvroRuntimeException(e);
+        } catch (MalformedURLException e) {
+            throw new AvroRuntimeException(e);
+        } catch (URISyntaxException e) {
+            throw new AvroRuntimeException(e);
+        }
     }
 
     public com.linkedin.avro.fastserde.generated.avro.StringableRecord deserializeStringableRecord0(Object reuse, Decoder decoder)
-        throws IOException
+        throws IOException, NumberFormatException, MalformedURLException, URISyntaxException
     {
         com.linkedin.avro.fastserde.generated.avro.StringableRecord StringableRecord;
         if ((reuse)!= null) {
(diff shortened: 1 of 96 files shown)


avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/serialization/AVRO_1_4/StringableRecord_SpecificSerializer_6174384286732341990.java
```diff
@@ -88,6 +88,20 @@ public void serializeStringableRecord0(StringableRecord data, Encoder encoder)
         serializeStringableSubRecord0(subRecord0, (encoder));
         AnotherSubRecord subRecordWithSubRecord0 = ((AnotherSubRecord) data.get(8));
         serializeAnotherSubRecord0(subRecordWithSubRecord0, (encoder));
+        CharSequence stringUnion0 = ((CharSequence) data.get(9));
+        if (stringUnion0 == null) {
+            (encoder).writeIndex(0);
+            (encoder).writeNull();
+        } else {
+            if (stringUnion0 instanceof CharSequence) {
```
FelixGV, line 96: Why do we have this `instanceof` check? It should always be true, right? Can we get rid of it?

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/serialization/AVRO_1_8/StringableRecord_SpecificSerializer_6174384286732341990.java
```diff
@@ -62,6 +62,16 @@ public void serializeStringableRecord0(StringableRecord data, Encoder encoder)
         serializeStringableSubRecord0(subRecord0, (encoder));
         AnotherSubRecord subRecordWithSubRecord0 = ((AnotherSubRecord) data.get(8));
         serializeAnotherSubRecord0(subRecordWithSubRecord0, (encoder));
+        String stringUnion0 = ((String) data.get(9));
+        if (stringUnion0 == null) {
+            (encoder).writeIndex(0);
+            (encoder).writeNull();
+        } else {
+            if (stringUnion0 instanceof String) {
```
FelixGV, line 70: Unnecessary check?

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_7/StringableRecord_SpecificDeserializer_6174384286732341990_6174384286732341990.java
```diff
@@ -107,6 +107,13 @@ public StringableRecord_SpecificDeserializer_1822942669987229249_182294266998722
         StringableRecord.put(6, urlMap0);
         StringableRecord.put(7, deserializeStringableSubRecord0(StringableRecord.get(7), (decoder)));
         StringableRecord.put(8, deserializeAnotherSubRecord0(StringableRecord.get(8), (decoder)));
+        int unionIndex0 = (decoder.readIndex());
+        if (unionIndex0 == 0) {
+            decoder.readNull();
+        }
+        if (unionIndex0 == 1) {
```
FelixGV, line 114: Can this be an else if, since it should never be useful to evaluate it if the previous condition was true?

avro-fastserde/src/test/java/com/linkedin/avro/fastserde/FastStringableTest.java
```diff
@@ -200,14 +203,15 @@ public void deserializeStringableFields(Boolean whetherUseFastDeserializer)
...
 
-    StringableRecord record = generateRecord(exampleURL, exampleURI, exampleFile, exampleBigDecimal, exampleBigInteger);
+    StringableRecord record = generateRecord(exampleURL, exampleURI, exampleFile, exampleBigDecimal, exampleBigInteger, exampleString);
 
     // when
     StringableRecord afterDecoding = null;
     if (whetherUseFastDeserializer) {
       afterDecoding =
-          readWithFastAvro(StringableRecord.SCHEMA$, StringableRecord.SCHEMA$, specificDataAsDecoder(record), true);
+          readWithFastAvro(StringableRecord.SCHEMA$, StringableRecord.SCHEMA$, writeWithFastAvro(record, StringableRecord.SCHEMA$, true), true);
```
FelixGV, line 214: It would be nice if the test change was in a separate commit, since we cannot see what changed in the generated code... We only see that there is new generated code, but not what the previous state was.

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/serialization/AVRO_1_4/StringableRecord_SpecificSerializer_6174384286732341990.java
gaojieliu: It  is not necessary..
I noticed this inefficiency, but I didn't fix it because of laziness..
Thanks for pushing this optimization.

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/serialization/AVRO_1_8/StringableRecord_SpecificSerializer_6174384286732341990.java
gaojieliu: It  is not necessary..
I noticed this inefficiency, but I didn't fix it because of laziness..
Thanks for pushing this optimization.

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_7/StringableRecord_SpecificDeserializer_6174384286732341990_6174384286732341990.java
gaojieliu: Fixed.

avro-fastserde/src/test/java/com/linkedin/avro/fastserde/FastStringableTest.java
gaojieliu: If you look at the diff side-by-side, you should be able to figure out what changes with the test change.
github treats those generated files as moved, so the diff is clear to me.
Let us know if it doesn't work for you.

FelixGV: So are you saying that the test change resulted in zero change to the generated code? I thought the test added a new field to the schema, which should change the generated code. Otherwise, how can we see how the generated code is different, if the change touches only a part which was previously not getting tested/generated?

gaojieliu: I got your point.
The avro schema change did result in changes in the generated code for both serializer and de-serializer, and the change to the schema is to verify the fix.
Without the latest code gen logic, the generated fast serializer couldn't compile since it would generate something like this:
```
String stringUnion0 = ((String) data.get(9));
        if (stringUnion0 == null) {
            (encoder).writeIndex(0);
            (encoder).writeNull();
        } else {
            (encoder).writeIndex(1);
            if (stringUnion0 instanceof Utf8) {
                (encoder).writeString(((Utf8) stringUnion0));
            } else {
                (encoder).writeString(stringUnion0 .toString());
            }
        }
```
If we want to  compare the difference before/after code gen change with the change of test schema, we need to checkin the invalid java file, it could be fine since it is temporary, but I am wondering whether it is mandatory for reviewing.
As long as the new part of the generated code is reasonable for the new change to the test schema, it should be fine, right?

gaojieliu:
Thanks @FelixGV for the review!

PR #70, opened 2020-06-30
"avro.java.string" support in union type for fast serializer

gaojieliu:
@FelixGV @volauvent @radai-rosenblatt

PR #72, opened 2020-07-03
fix for 'avro.java.string' property handling when set on reader schema

flowenol:
* 'avro.java.string' properly handled when set on reader schema
* 'avro.java.string' properly handled when set on map key


flowenol:
Hi,

'avro.java.string' property was improperly handled when set only on reader schema upon deserialization, so I prepared fix based on the original RTBHOUSE/avro-fastserde repository implementation. Moreover, the same property on map keys was not handled at all, so I added that capability as well.

Cheers,
Piotr

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -539,7 +544,12 @@ private void processUnion(JVar unionSchemaVar, final String name, final Schema u
         processComplexType(optionSchemaVar, optionName, optionSchema, readerOptionSchema, thenBlock, unionAction,
             putValueIntoParent, reuseSupplier);
       } else {
-        processSimpleType(optionSchema, thenBlock, unionAction, putValueIntoParent, reuseSupplier);
+        // to preserve string specific options use reader option schema
+        if (action.getShouldRead() && Schema.Type.STRING.equals(optionSchema.getType())) {
```
gaojieliu, line 548: I could see repeatable code in this PR, so could you push down the common logic to function `processSimpleType` and pass both reader schema and write schema to function: `processSimpleType`?

avro-fastserde/src/test/java/com/linkedin/avro/fastserde/FastStringableTest.java
```diff
@@ -143,11 +147,24 @@ public void serializeStringableFields() throws URISyntaxException, MalformedURLE
   }
 
   @Test(groups = {"deserializationTest"}, dataProvider = "SlowFastDeserializer")
-  public void javaStringPropertyTest(Boolean whetherUseFastDeserializer) throws IOException {
-    Schema schema = createRecord(createField("testString", javaStringSchema));
+  @SuppressWarnings("unchecked")
+  public void javaStringPropertyTest(Boolean whetherUseFastDeserializer) {
```
gaojieliu, line 151: Could you commit  the changed files in codegen directory as well? so that we could see the differences before/after this code gen logic change.
Sample PR:
https://github.com/linkedin/avro-util/pull/70/files

Essentially, we would split the code change into  two parts:
1. core code change;
2. the generated fast serializer/de-serializer change;

With this way, we will be able to review those two set of changes separately by choosing  different commits.

For the PR link I shared, it  contains the following two commits:
"avro.java.string" support in union type for fast serializer

@gaojieliu
gaojieliu committed 13 days ago
 
Code gen for #70

@gaojieliu
gaojieliu committed 13 days ago


Let me know if you are still not clear about this new implicit convention formed inside LinkedIn.

flowenol: Hi,

Thank you for clarification, the PR has been changed according to your suggestions.

gaojieliu:
@flowenol 
Thanks for the fix.
The code change has been merged.
Do you need your another PR to be committed before cutting  a new release?
https://github.com/linkedin/avro-util/pull/75

flowenol:
@gaojieliu 

Thank you for accepting this PR. #75 would be nice to have as well.

PR #75, opened 2020-07-07
logger verbosity changes

flowenol:
* schema content is not logged upon successful fast(de)serializer compilation by default
* success messages are logged after actual (de)serializer generation in FastSerdeCache

### DIFF
--- a/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
+++ b/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
@@ -1,5 +1,8 @@
 package com.linkedin.avro.fastserde;
 
+import static com.linkedin.avro.fastserde.Utils.getSchemaFingerprint;
+import static com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper.getSchemaFullName;
+
 import java.io.File;
 import java.io.IOException;
 import java.lang.reflect.ParameterizedType;
@@ -15,6 +18,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
+
 import org.apache.avro.Schema;
 import org.apache.avro.generic.ColdGenericDatumReader;
 import org.apache.avro.generic.ColdSpecificDatumReader;
@@ -292,8 +296,8 @@ public final class FastSerdeCache {
   }
 
   private String getSchemaKey(Schema writerSchema, Schema readerSchema) {
-    return String.valueOf(Math.abs(Utils.getSchemaFingerprint(writerSchema))) + Math.abs(
-        Utils.getSchemaFingerprint(readerSchema));
+    return String.valueOf(Math.abs(getSchemaFingerprint(writerSchema))) + Math.abs(
+        getSchemaFingerprint(readerSchema));
   }
 
   /**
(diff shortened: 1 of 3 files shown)


flowenol:
Hi,

The rationale behind this change is to reduce the default logging verbosity, especially by elimination of schema content in logging messages. In our scenario this causes unnecessary pollution of logs, due to the massive schemas used in our scenarios. 

This change only eliminates schema logging upon successful generation of fast(de)serializers. In case of failure full schema content will be available as before.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -308,9 +310,20 @@ private String getSchemaKey(Schema writerSchema, Schema readerSchema) {
...
+    FastDeserializer<?> fastDeserializer = generator.generateDeserializer();
+
+    if (LOGGER.isDebugEnabled()) {
+      LOGGER.debug("Generated classes dir: {} and generation of specific FastDeserializer is done for writer schema of type: {} with fingerprint: {}"
+              + " and content: [\n{}\n] and reader schema of type: {} with fingerprint: {} and content: [\n{}\n]", classesDir, writerSchema.getFullName(),
+              writerSchema.toString(true), getSchemaFingerprint(writerSchema), readerSchema.getFullName(), getSchemaFingerprint(readerSchema),
+              readerSchema.toString(true));
+    } else {
+      LOGGER.info("Generated classes dir: {} and generation of specific FastDeserializer is done for writer schema of type: {} with fingerprint: {}"
+              + " and reader schema of type: {} with fingerprint: {}", classesDir, writerSchema.getFullName(), getSchemaFingerprint(writerSchema),
```
radai-rosenblatt, line 322: getSchemaFingerprint() invokes Schema.hashCode() which (under avro 1.4) is not a cached value and hence slow. i think either ifInfoEnabled() or a decorator object who's toString() would evaluate to getSchemaFingerprint(schema) is warranted here

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -393,8 +418,16 @@ public Object deserialize(Object reuse, Decoder d) throws IOException {
...
-    LOGGER.info("Generated classes dir: {} and generation of specific FastSerializer is done for schema: [\n{}\n]",
-        classesDir, schema.toString(true));
+
+    if (LOGGER.isDebugEnabled()) {
+      LOGGER.debug("Generated classes dir: {} and generation of specific FastSerializer is done for schema of type: {}" +
+              " and fingerprint: {} and content: [\n{}\n]", classesDir, schema.getFullName(), getSchemaFingerprint(schema),
+              schema.toString(true));
+    } else {
+      LOGGER.info("Generated classes dir: {} and generation of specific FastSerializer is done for schema of type: {}" +
+              " and fingerprint: {}", classesDir, schema.getFullName(), getSchemaFingerprint(schema));
```
radai-rosenblatt, line 428: same concern about Scehema.hashCode() under avro 1.4 as above

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
volauvent: In Avro-1.4,  `getFullName()` is not supported if the Schema is not a NamedSchema.
I would suggest to print the complete schema here since it will be only printed for one time...

gaojieliu: Efficiency shouldn't be a  problem here since this log should only print once per schema pair.
Also I am wondering how valuable to print `fingerprint` here since we could tell a schema from a fingerprint.. `FullName` might be good enough..
If `FullName` is not supported, the `type` will be printed.

radai-rosenblatt:
i'll drop my perf concerns if this wont be called often. what about the avro 1.4 gullName concern though ?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
flowenol: Hi

As I have mentioned before, full schema printing is somewhat troublesome from our perspective. We use a lot of massive schemas operating in different versions what results in many logging messages spanning across multiple screens of text every time the successful compilation takes place. Aesthetically speaking I think that this is a little bit too verbose on INFO level. But feedback from the (de)serializers generation process is generally appreciated. I wanted to differentiate the logging messages for the particular versions of the same schema, so that is the reason why I used the fingerprint mechanism, however I wasn't aware of the potential performance issues in avro 1.4 which can be introduced this way. But as @gaojieliu observed, does this really pose a performance threat? In vast majority of cases this should be called only once at the beginning.

flowenol:
Regarding the avro 1.4 _getFullName_ incompatibility, how about introducing the helper method _getSchemaFullName_ in **AvroCompatibiltyHelper** class that would straighten this out by returning type in case of **AvroRuntimeException** from _getFullName_?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
radai-rosenblatt: i withdraw my performance concerns here

flowenol:
Hi,

Thank you for merging this PR. @radai-rosenblatt how about distributing a new release?

radai-rosenblatt:
@flowenol - just cut 0.2.7

PR #76, opened 2020-07-11
enable caching build artifacts across builds

radai-rosenblatt:


README.md
```diff
@@ -78,3 +78,7 @@ The helper module supports avro 1.4 - 1.9 inclusive. for fastserde support:
 | Avro 1.7 |      Yes      |      Yes        |      Yes           |      Yes             |
 | Avro 1.8 |      Yes      |      Yes        |      Yes           |      Yes             |
 | Avro 1.9 |      Yes      |      Yes        |      ???           |      ???             |
+
+## FAQ / common build issues ##
```
FelixGV, line 82: Can we rename this header "Contributors" or something like that? Should it even have its own separate page? It doesn't pertain to users, right?

radai-rosenblatt: created a DEVELOPER.md file

PR #80, opened 2020-07-21
add limit for cache

wxn225:
We set a hard limit for the cache based on our benchmark test.

### DIFF
--- a/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
+++ b/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
@@ -2,6 +2,8 @@ package com.linkedin.avro.fastserde;
 
 import java.io.File;
 import java.io.IOException;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Method;
 import java.lang.reflect.ParameterizedType;
 import java.net.URL;
 import java.net.URLClassLoader;
@@ -44,6 +46,21 @@ public final class FastSerdeCache {
 
   private static volatile FastSerdeCache _INSTANCE;
 
+  /**
+   * Fast-avro will generate and load serializer and deserializer(SerDes) classes into metaspace during runtime.
+   * During serialization and deserialization, fast-avro also leverages JIT compilation to boost the SerDes speed.
+   * And JIT compilation code is saved in code cache.
+   * Too much usage of metaspace and code cache will bring GC/OOM issue.
+   *
+   * We set a hard limit of the total number of SerDes classes generated and loaded by fast-avro.
+   * By default, the limit is set to MAX_INT.
+   * Fast-avro will fall back to regular avro after the limit is hit.
+   * One could set the limit through {@link FastSerdeCache} constructors.
+   * And get the limit by {@link #getGeneratedFastClassNumLimit}
+   */
+  private volatile int generatedFastClassNumLimit = Integer.MAX_VALUE;
+  private AtomicInteger schemaNum = new AtomicInteger(0);
+
   private final Map<String, FastDeserializer<?>> fastSpecificRecordDeserializersCache =
       new FastAvroConcurrentHashMap<>();
   private final Map<String, FastDeserializer<?>> fastGenericRecordDeserializersCache =
@@ -79,6 +96,20 @@ public final class FastSerdeCache {
    */
   public FastSerdeCache(Executor executorService, Supplier<String> compileClassPathSupplier) {
     this(executorService, compileClassPathSupplier.get());
+  }
+
+  /**
+   *
+   * @param executorService
+   *            {@link Executor} used by serializer/deserializer compile threads
+   * @param compileClassPathSupplier
+   *            custom classpath {@link Supplier}
+   * @param limit
+   *            custom number {@link #generatedFastClassNumLimit}
+   */
+  public FastSerdeCache(Executor executorService, Supplier<String> compileClassPathSupplier, int limit) {
+    this(executorService, compileClassPathSupplier);
+    setGeneratedFastClassNumLimit(limit);
   }
 
   public FastSerdeCache(String compileClassPath) {
@@ -120,6 +151,15 @@ public final class FastSerdeCache {
 
   private FastSerdeCache() {
     this((Executor) null);
+  }
+
+  /**
+   * @param limit
+   *            custom number {@link #generatedFastClassNumLimit}
+   */
+  public FastSerdeCache(int limit) {
+    this();
+    setGeneratedFastClassNumLimit(limit);
   }
 
   /**
@@ -198,22 +238,8 @@ public final class FastSerdeCache {
    * @return specific-class aware avro {@link FastDeserializer}
    */
   public FastDeserializer<?> getFastSpecificDeserializer(Schema writerSchema, Schema readerSchema) {
-    String schemaKey = getSchemaKey(writerSchema, readerSchema);
-    FastDeserializer<?> deserializer = fastSpecificRecordDeserializersCache.get(schemaKey);
-
-    if (deserializer == null) {
-      deserializer = fastSpecificRecordDeserializersCache.putIfAbsent(schemaKey,
-          new FastDeserializerWithAvroSpecificImpl<>(writerSchema, readerSchema));
-      if (deserializer == null) {
-        deserializer = fastSpecificRecordDeserializersCache.get(schemaKey);
-        CompletableFuture.supplyAsync(() -> buildSpecificDeserializer(writerSchema, readerSchema), executor)
-            .thenAccept(d -> {
-              fastSpecificRecordDeserializersCache.put(schemaKey, d);
-            });
-      }
-    }
-
-    return deserializer;
+    return getDeSerializer(writerSchema, readerSchema, fastSpecificRecordDeserializersCache,
+        FastDeserializerWithAvroSpecificImpl.class, "buildSpecificDeserializer");
   }
 
   /**
@@ -226,21 +252,8 @@ public final class FastSerdeCache {
    * @return generic-class aware avro {@link FastDeserializer}
    */
   public FastDeserializer<?> getFastGenericDeserializer(Schema writerSchema, Schema readerSchema) {
-    String schemaKey = getSchemaKey(writerSchema, readerSchema);
-    FastDeserializer<?> deserializer = fastGenericRecordDeserializersCache.get(schemaKey);
-
-    if (deserializer == null) {
-      deserializer = fastGenericRecordDeserializersCache.putIfAbsent(schemaKey,
-          new FastDeserializerWithAvroGenericImpl(writerSchema, readerSchema));
-      if (deserializer == null) {
-        deserializer = fastGenericRecordDeserializersCache.get(schemaKey);
-        CompletableFuture.supplyAsync(() -> buildGenericDeserializer(writerSchema, readerSchema), executor)
-            .thenAccept(d -> {
-              fastGenericRecordDeserializersCache.put(schemaKey, d);
-            });
-      }
-    }
-    return deserializer;
+    return getDeSerializer(writerSchema, readerSchema, fastGenericRecordDeserializersCache,
+        FastDeserializerWithAvroGenericImpl.class, "buildFastGenericDeserializer");
   }
 
   /**
@@ -251,20 +264,8 @@ public final class FastSerdeCache {
    * @return specific-class aware avro {@link FastSerializer}
    */
   public FastSerializer<?> getFastSpecificSerializer(Schema schema) {
-    String schemaKey = getSchemaKey(schema, schema);
-    FastSerializer<?> serializer = fastSpecificRecordSerializersCache.get(schemaKey);
-    if (serializer == null) {
-      serializer =
-          fastSpecificRecordSerializersCache.putIfAbsent(schemaKey, new FastSerializerWithAvroSpecificImpl(schema));
-      if (serializer == null) {
-        serializer = fastSpecificRecordSerializersCache.get(schemaKey);
-        CompletableFuture.supplyAsync(() -> buildSpecificSerializer(schema), executor).thenAccept(s -> {
-          fastSpecificRecordSerializersCache.put(schemaKey, s);
-        });
-      }
-    }
-
-    return serializer;
+    return getDeSerializer(schema, schema, fastSpecificRecordSerializersCache,
+        FastSerializerWithAvroSpecificImpl.class, "buildSpecificSerializer");
   }
 
   /**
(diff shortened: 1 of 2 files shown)


avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -44,6 +44,22 @@
...
+   * During serialization and deserialization, fast-avro also leverages JIT compilation to boost the SerDes speed.
+   * And JIT compilation code is saved in code cache.
+   * Too much usage of metaspace and code cache will bring GC/OOM issue.
+   *
+   * We set a hard limit of the total number of SerDes classes generated and loaded by fast-avro.
+   * Currently we set the limit to 400 based on our performance test with manually generated schemas.
+   * The number could keep the metaspace and codecache usage within a safe range.
+   * Fast-avro will fall back to regular avro after the limit is hit.
+   *
+   * TODO: Let users configure the limit.
```
FelixGV, line 58: I don't think we should merge this in before making it configurable...

volauvent:
Hi @wxn225 Please finish the following steps to make the PR ready for reviewing.
1. Add test and comments for the new functions you have added.
2. Squash commits into one.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
wxn225: Hi, I added set/get limit functions in the cache. And also added getFastSerdeCache function in datum reader/writer. So now it is configurable.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -206,10 +221,14 @@ public static boolean isSupportedForFastSerializer(Schema.Type schemaType) {
           new FastDeserializerWithAvroSpecificImpl<>(writerSchema, readerSchema));
       if (deserializer == null) {
         deserializer = fastSpecificRecordDeserializersCache.get(schemaKey);
-        CompletableFuture.supplyAsync(() -> buildSpecificDeserializer(writerSchema, readerSchema), executor)
-            .thenAccept(d -> {
-              fastSpecificRecordDeserializersCache.put(schemaKey, d);
-            });
+        if (schemaNum.get() < schemaNumLimit) {
```
gaojieliu, line 224: I could this piece of code is duplicate in several different functions.
Can we abstract a function to wrap up the common logic to enforce the limit check?

For example, you could pass the actual generation logic as a runnable.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -44,6 +44,21 @@
...
+   * And JIT compilation code is saved in code cache.
+   * Too much usage of metaspace and code cache will bring GC/OOM issue.
+   *
+   * We set a hard limit of the total number of SerDes classes generated and loaded by fast-avro.
+   * By default, the limit is set to 400 based on our performance test with manually generated schemas.
+   * The number could keep the metaspace and codecache usage within a safe range.
+   * Fast-avro will fall back to regular avro after the limit is hit.
+   * One could get/set the limit by {@link #getSchemaNumLimit} and {@link #setSchemaNumLimit}
+   */
+  private static volatile int schemaNumLimit = 400;
```
gaojieliu, line 59: From my understanding, the goal is to limit the total number of the generated classes, so could we rename it to be `generatedFastClassNumLimit`?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -296,6 +329,24 @@ private String getSchemaKey(Schema writerSchema, Schema readerSchema) {
...
+  public int getSchemaNumLimit() {
+    return schemaNumLimit;
+  }
+
+  /**
+   * Set the total schema number limit for {@link FastSerdeCache}.
+   * It is the total limit for fast Generic/Specific De/Serializer
+   * If the limit is set for multiple times, the last limit will be used
+   */
+  public void setSchemaNumLimit(int limit) {
```
gaojieliu, line 346: I think it is better to pass the `limit` as the param of a new constructor and it shouldn't change during the lifetime of `FastSerdeCache`.
If we choose to expose a `setter` here, when the user specifies a lower number after generating a lot of fast classes, fast-avro won't unload the loaded classes to satisfy the new limit, which is a little confusing.

And I don't see a reason people would change the limit dynamically since they have no notion about how many classes have been generated or how many more are needed in the future when the application is running.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
wxn225: Yes, I'm working on that.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
radai-rosenblatt: also - can we default this to max int? it would be better if people opted into this limit than if they find this out possibly weeks after deployment

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -206,10 +221,14 @@ public static boolean isSupportedForFastSerializer(Schema.Type schemaType) {
...
-        CompletableFuture.supplyAsync(() -> buildSpecificDeserializer(writerSchema, readerSchema), executor)
-            .thenAccept(d -> {
-              fastSpecificRecordDeserializersCache.put(schemaKey, d);
-            });
+        if (schemaNum.get() < schemaNumLimit) {
+          CompletableFuture.supplyAsync(() -> buildSpecificDeserializer(writerSchema, readerSchema), executor).thenAccept(d -> {
+            fastSpecificRecordDeserializersCache.put(schemaKey, d);
+          });
+        } else if (schemaNum.get() == schemaNumLimit) {
+          LOGGER.info("Generated fast de/serializer number hits " + schemaNumLimit + " limit");
```
radai-rosenblatt, line 229: make this a warning. this is definitely something oenwers of running code should pay attention to

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -234,10 +253,14 @@ public static boolean isSupportedForFastSerializer(Schema.Type schemaType) {
...
-        CompletableFuture.supplyAsync(() -> buildGenericDeserializer(writerSchema, readerSchema), executor)
-            .thenAccept(d -> {
-              fastGenericRecordDeserializersCache.put(schemaKey, d);
-            });
+        if(schemaNum.get() < schemaNumLimit) {
+          CompletableFuture.supplyAsync(() -> buildGenericDeserializer(writerSchema, readerSchema), executor).thenAccept(d -> {
+            fastGenericRecordDeserializersCache.put(schemaKey, d);
+          });
+        } else if (schemaNum.get() == schemaNumLimit) {
+          LOGGER.info("Generated fast de/serializer number hits " + schemaNumLimit + " limit");
```
radai-rosenblatt, line 261: warn

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
radai-rosenblatt: also, can you print the value of the counter as well as the max? without it users wont have a clue what they should set the limit to

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -258,9 +281,14 @@ public static boolean isSupportedForFastSerializer(Schema.Type schemaType) {
...
         serializer = fastSpecificRecordSerializersCache.get(schemaKey);
-        CompletableFuture.supplyAsync(() -> buildSpecificSerializer(schema), executor).thenAccept(s -> {
-          fastSpecificRecordSerializersCache.put(schemaKey, s);
-        });
+        if (schemaNum.get() < schemaNumLimit) {
+          CompletableFuture.supplyAsync(() -> buildSpecificSerializer(schema), executor).thenAccept(s -> {
+            fastSpecificRecordSerializersCache.put(schemaKey, s);
+          });
+        } else if (schemaNum.get() == schemaNumLimit) {
+          LOGGER.info("Generated fast de/serializer number hits " + schemaNumLimit + " limit");
```
radai-rosenblatt, line 289: warn, print counter

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -296,6 +329,24 @@ private String getSchemaKey(Schema writerSchema, Schema readerSchema) {
         Utils.getSchemaFingerprint(readerSchema));
   }
 
+  /**
+   * This function can get the total schema number limit for {@link FastSerdeCache}
+   *
+   * @return current schema number limit
+   */
+  public int getSchemaNumLimit() {
+    return schemaNumLimit;
```
radai-rosenblatt, line 338: schemaNumLimit is static. why is this method not static?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
radai-rosenblatt: 1. schemaNumLimit is static - so this method should be static as well.
2. that means there's no constructor to set this in?
3. i would expect this would be called once, during application boot. however i see no damage is changing this at runtime

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -44,6 +46,21 @@
...
+   * Too much usage of metaspace and code cache will bring GC/OOM issue.
+   *
+   * We set a hard limit of the total number of SerDes classes generated and loaded by fast-avro.
+   * By default, the limit is set to MAX_INT.
+   * Fast-avro will fall back to regular avro after the limit is hit.
+   * One could set the limit through {@link FastSerdeCache} constructors.
+   * And get the limit by {@link #getGeneratedFastClassNumLimit}
+   */
+  private volatile int generatedFastClassNumLimit = Integer.MAX_VALUE;
+  private AtomicInteger schemaNum = new AtomicInteger(0);
```
radai-rosenblatt, line 62: should be final

wxn225: schemaNum to be final?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -275,27 +276,84 @@ public static boolean isSupportedForFastSerializer(Schema.Type schemaType) {
...
+   * @return a fast specific/generic de/serializer
+   */
+  private <T> T getDeSerializer(Schema writerSchema, Schema readerSchema, Map<String, T> cache, Class<?> avroImplClass,
+      String buildFastDeSerializerFuncName){
+    String schemaKey = getSchemaKey(writerSchema, readerSchema);
+    T deserializer = cache.get(schemaKey);
+    if (deserializer == null) {
+      try {
+        Constructor cs = avroImplClass.getConstructor(Schema.class, Schema.class);
+        deserializer = cache.putIfAbsent(schemaKey, (T)cs.newInstance(writerSchema, readerSchema));
```
radai-rosenblatt, line 323: putIfAbsent 1st constructs the argument (calling cs) and only then checks if the map already has a value.
computeIfAbsent would allow you to defer calling cs and hence would be slightly more efficient

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -275,27 +276,84 @@ public static boolean isSupportedForFastSerializer(Schema.Type schemaType) {
...
+      } catch (Exception e) {
+        e.printStackTrace();
```
radai-rosenblatt, line 325: this class has a logger - you can print an error (exception included)

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -275,27 +276,84 @@ public static boolean isSupportedForFastSerializer(Schema.Type schemaType) {
...
+        deserializer = cache.get(schemaKey);
+        if (schemaNum.get() < generatedFastClassNumLimit) {
+          try {
+            Method
+                buildDeSerializer = FastSerdeCache.class.getMethod(buildFastDeSerializerFuncName, Schema.class, Schema.class);
+            CompletableFuture.supplyAsync(() -> {
+              try {
+                return (T)buildDeSerializer.invoke(_INSTANCE ,writerSchema, readerSchema);
+              } catch (Exception e) {
+                e.printStackTrace();
```
radai-rosenblatt, line 337: log error

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -275,27 +276,84 @@ public static boolean isSupportedForFastSerializer(Schema.Type schemaType) {
...
+              }
+              return null;
+            }, executor).thenAccept(s -> {
+              cache.put(schemaKey, s);
+            });
+          }
+          catch (Exception e) {
+            e.printStackTrace();
```
radai-rosenblatt, line 345: log error

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -275,27 +276,84 @@ public static boolean isSupportedForFastSerializer(Schema.Type schemaType) {
...
+    if (deserializer == null) {
+      try {
+        Constructor cs = avroImplClass.getConstructor(Schema.class, Schema.class);
+        deserializer = cache.putIfAbsent(schemaKey, (T)cs.newInstance(writerSchema, readerSchema));
+      } catch (Exception e) {
+        e.printStackTrace();
+      }
+      if (deserializer == null) {
+        deserializer = cache.get(schemaKey);
+        if (schemaNum.get() < generatedFastClassNumLimit) {
```
radai-rosenblatt, line 329: moving the incrementAndGet call here (instead of a simple get) is more correct

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -398,7 +456,7 @@ public Object deserialize(Object reuse, Decoder d) throws IOException {
     return generator.generateSerializer();
   }
 
-  private FastSerializer<?> buildSpecificSerializer(Schema schema) {
+  private FastSerializer<?> buildSpecificSerializer(Schema schema, Schema uselessSchema) {
```
radai-rosenblatt, line 459: why do we need this 2nd arg?

PR #81, opened 2020-07-22
Added specific record tests for primitive arrays.

FelixGV:
Addressed bugs related to missing constructors in the PrimitiveArrayList
and its child classes. Also refactored those classes slightly to reduce
boilerplate.

Altered the meta code:

- List construction does not rely on Collections.emptyList() since that
  is incompatible with the primitive list classes.
- Added a missing getShouldRead() check in cases where schema evolution
  gets rid of a list field. This would previously cause a NPE during
  code gen.

### DIFF
--- a/avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/StringableRecord_SpecificDeserializer_6174384286732341990_6174384286732341990.java
+++ b/avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/StringableRecord_SpecificDeserializer_6174384286732341990_6174384286732341990.java
@@ -65,29 +65,25 @@ public class StringableRecord_SpecificDeserializer_6174384286732341990_617438428
         }
         List<Utf8> urlArray0 = null;
         long chunkLen0 = (decoder.readArrayStart());
-        if (chunkLen0 > 0) {
-            if (StringableRecord.get(5) instanceof List) {
-                urlArray0 = ((List) StringableRecord.get(5));
-                urlArray0 .clear();
-            } else {
-                urlArray0 = new ArrayList<Utf8>();
-            }
-            do {
-                for (int counter0 = 0; (counter0 <chunkLen0); counter0 ++) {
-                    Object urlArrayArrayElementReuseVar0 = null;
-                    if (StringableRecord.get(5) instanceof GenericArray) {
-                        urlArrayArrayElementReuseVar0 = ((GenericArray) StringableRecord.get(5)).peek();
-                    }
-                    if (urlArrayArrayElementReuseVar0 instanceof Utf8) {
-                        urlArray0 .add((decoder).readString(((Utf8) urlArrayArrayElementReuseVar0)));
-                    } else {
-                        urlArray0 .add((decoder).readString(null));
-                    }
-                }
-                chunkLen0 = (decoder.arrayNext());
-            } while (chunkLen0 > 0);
+        if (StringableRecord.get(5) instanceof List) {
+            urlArray0 = ((List) StringableRecord.get(5));
+            urlArray0 .clear();
         } else {
-            urlArray0 = Collections.emptyList();
+            urlArray0 = new ArrayList<Utf8>();
+        }
+        while (chunkLen0 > 0) {
+            for (int counter0 = 0; (counter0 <chunkLen0); counter0 ++) {
+                Object urlArrayArrayElementReuseVar0 = null;
+                if (StringableRecord.get(5) instanceof GenericArray) {
+                    urlArrayArrayElementReuseVar0 = ((GenericArray) StringableRecord.get(5)).peek();
+                }
+                if (urlArrayArrayElementReuseVar0 instanceof Utf8) {
+                    urlArray0 .add((decoder).readString(((Utf8) urlArrayArrayElementReuseVar0)));
+                } else {
+                    urlArray0 .add((decoder).readString(null));
+                }
+            }
+            chunkLen0 = (decoder.arrayNext());
         }
         StringableRecord.put(5, urlArray0);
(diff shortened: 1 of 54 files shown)


avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/StringableRecord_SpecificDeserializer_6174384286732341990_6174384286732341990.java
```diff
@@ -87,7 +87,7 @@ public StringableRecord_SpecificDeserializer_6174384286732341990_617438428673234
                 chunkLen0 = (decoder.arrayNext());
             } while (chunkLen0 > 0);
         } else {
-            urlArray0 = Collections.emptyList();
+            urlArray0 = new ArrayList<Utf8>();
```
radai-rosenblatt, line 90: nit - initial size 1?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -144,6 +146,9 @@
     } catch (JClassAlreadyExistsException e) {
       throw new FastDeserializerGeneratorException("Class: " + className + " already exists");
     } catch (Exception e) {
+      LOGGER.error("Failed to generate a deserializer.");
```
radai-rosenblatt, line 149: better to have this as a single line (maybe with \n) because in a busy system other prints can get injected in the middle on the log file

radai-rosenblatt: also - use {} and dont call toString() (less important for error, but still)

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -662,10 +667,8 @@ private void processArray(JVar arraySchemaVar, final String name, final Schema a
       }
       doLoop.body().assign(chunkLen, JExpr.direct(DECODER + ".arrayNext()"));
     }, else1 -> {
-      if (useGenericTypes) {
+      if (finalAction.getShouldRead()) {
         else1.assign(arrayVar, finalNewArrayExp);
```
radai-rosenblatt, line 671: this generates an arraylist of size 10, i think it should make it smaller (since its always empty here)

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/StringableRecord_SpecificDeserializer_6174384286732341990_6174384286732341990.java
volauvent: why not 0 since it's always an empty ArrayList? :)

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
volauvent: We do not need to log here since the `FastDeserializerGeneratorException` would be catched in upper layer `buildGeneric/SpecificDeserializer` and schemas would be logged?

avro-fastserde/src/test/avro/defaultsTest.avsc
```diff
@@ -350,6 +350,54 @@
         }
       },
       "default": {"test": [null]}
+    },
+    {
```
volauvent, line 354: No code-gen results for these new fields in this PR?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
FelixGV: Hmm... I think the thing is that we have test code that invokes this function directly, and therefore does not benefit from the upstream logging that happens in the main code. I will remove those extra logs then.

avro-fastserde/src/test/avro/defaultsTest.avsc
FelixGV: No, we currently don't add this to the codegen directory. Only some of the tests are put in there. Perhaps those should be added...

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/StringableRecord_SpecificDeserializer_6174384286732341990_6174384286732341990.java
FelixGV: I don't really like the structure of that generated code, and I was meaning to tweak it anyway at some point... so might as well do it now.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -606,68 +609,61 @@ private void processArray(JVar arraySchemaVar, final String name, final Schema a
     }
     JInvocation finalNewArrayExp = newArrayExp;
 
-    JExpression chunkLengthGreaterThanZero = chunkLen.gt(JExpr.lit(0));
 
-    ifCodeGen(parentBody, chunkLengthGreaterThanZero, then1 -> {
```
FelixGV, line 611: FYI: this looks like a big change but it's mostly just a move and an indentation change. The list instantiation (and potential reuse) now happens first, and the `if chunk length > 0` check as well as its else block is now removed. Then most of the remaining code is just an indentation change...

PR #82, opened 2020-07-24
add hard limit for FastSerdeCache

wxn225:


volauvent:
LGMT. Thanks

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -296,6 +339,26 @@ private String getSchemaKey(Schema writerSchema, Schema readerSchema) {
...
+      if (this.generatedSerDesNum.get() < this.generatedFastSerDesLimit) {
+        runnable.run();
+      } else if (this.generatedSerDesNum.get() == this.generatedFastSerDesLimit) {
+        LOGGER.warn("Generated fast serdes classes number hits limit {}", this.generatedFastSerDesLimit);
+      } else if (this.generatedSerDesNum.get() > this.generatedFastSerDesLimit) {
+        LOGGER.debug("Generated serdes number {}, with fast serdes limit set to {}", this.generatedSerDesNum.get(), this.generatedFastSerDesLimit);
+      }
+      this.generatedSerDesNum.incrementAndGet();
+    } catch (Exception e) {
+      LOGGER.error("Fast serdes class generation failed");
```
gaojieliu, line 358: Shall we print out the full error stacktrace here?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -296,6 +339,26 @@ private String getSchemaKey(Schema writerSchema, Schema readerSchema) {
...
+  private void buildFastClassWithNumCheck(Runnable runnable) {
+    try {
+      if (this.generatedSerDesNum.get() < this.generatedFastSerDesLimit) {
+        runnable.run();
+      } else if (this.generatedSerDesNum.get() == this.generatedFastSerDesLimit) {
+        LOGGER.warn("Generated fast serdes classes number hits limit {}", this.generatedFastSerDesLimit);
+      } else if (this.generatedSerDesNum.get() > this.generatedFastSerDesLimit) {
+        LOGGER.debug("Generated serdes number {}, with fast serdes limit set to {}", this.generatedSerDesNum.get(), this.generatedFastSerDesLimit);
+      }
+      this.generatedSerDesNum.incrementAndGet();
```
gaojieliu, line 356: Race condition could happen here since the check and update are happening in different places.
I don't think it is a big concern considering the default thread pool size is 2, so in the worst scenario, the total number of the generated classes could be one more than the specified limit.
But it is good to  call it out here, and we need to use `AtomicInteger` here since there is no difference from an `Integer` according to the actual usage.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
wxn225: The exceptions are handled in other functions. So a stack trace here may not provide any useful information.

PR #83, opened 2020-08-05
Added a config to limit the code-gen and class loading

volauvent:
The config sets the maximum number of fast SerDes classes
generated and loaded. It helps to limit the metaspace and codecache
usage brought by fast-avro runtime code-gen and class loading.

The limit config can be set via FastSerdeCache constructors.


avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeBase.java
```diff
@@ -136,6 +146,29 @@ protected Class compileClass(final String className, Set<String> knownUsedFullyQ
...
+  protected synchronized <T> T loadClassWithLimit(Supplier<T> supplier) {
+    if (this.loadClassNum < this.loadClassLimit) {
+      T classObj = null;
+      classObj = supplier.get();
+      this.loadClassNum++;
+      return classObj;
+    } else {
+      LOGGER.warn("Loaded fast serdes classes number {}, with limit set to {}", loadClassNum, loadClassLimit);
+    }
+    return null;
```
gaojieliu, line 172: I think it will be clearer to just throw the exception instead of  returning null.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -122,10 +150,24 @@ public FastSerdeCache(Executor executorService) {
...
+
   private FastSerdeCache() {
     this((Executor) null);
   }
 
+  /**
+   * @param limit
+   *            custom number {@link #generatedFastSerDesLimit}
+   */
+  public FastSerdeCache(int limit) {
```
gaojieliu, line 166: Do we need to add other two constructors? Maybe this one is good enough.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -300,6 +342,25 @@ private String getSchemaKey(Schema writerSchema, Schema readerSchema) {
...
+  /**
+   * A wrapper function that limits the total number of fast de/serializer classes generated
+   *
+   * @param supplier The function to build and save fast de/serializer
+   */
+  private <T> T buildFastClassWithLimit(Supplier<T> supplier) {
+    T result = null;
+    if (this.generatedSerDesNum.get() < this.generatedFastSerDesLimit) {
+      result = supplier.get();
+    } else if (this.generatedSerDesNum.get() == this.generatedFastSerDesLimit) {
```
gaojieliu, line 354: I am seeing there are two counters, which are being maintained independently in FastSerdeCache and FastSerdeBase, and I think we should combine and just use one.
The idea in my mind:
1. Maintaining the counter and limit in FastSerdeCache.
2. Passing the limit enforcement function to FastSerdeBase via all the Generator classes.
3. The limit enforcement function can be this way:
```
private int generatedSerDeNum;
  private int generatedFastSerDeLimit;
  Predicate<Boolean> limitPredicate = (whetherIncrementCounter) -> {
    synchronized (this) {
      if (generatedSerDeNum >= generatedFastSerDeLimit) {
        return false;
      }
      if (whetherIncrementCounter) {
        ++generatedSerDeNum;
      }
      return true;
    }
  };
```
4. In FastSerdeCache and all kinds of generators before the fast class generation, we should call `limitPredicate(false)`, and fail fast when this predicate returns false.
5.  In FastSerdeBase, before loading the new generated class, we should call `limitPredicate(true)`, and fail fast if the predicate return false.

Essentially, the idea is to keep the counting  logic in a single place to make it consistent.
Invoking this function in FastSerdeCache and various Generators is to try to avoid useless work.

Let me know if you want to have a sync up about this.

volauvent: Thanks for the detailed explaination!

This approach indeed helps to reduce two counters to one. However, it also leads to redundant code-gens and compilations which are CPU intensive tasks. IIUC, there's no limit for the code-gen and compilation before we really load  `generatedFastSerDeLimit` number of fast classes. So the corner case is fast-avro may generate and compile a great number of fast classes, and then throw them away.

For the current implementation, we at most generate extra N - 1 fast classes, N is the threads number of FastSerdeCache Executor. So I think the current implementation is better. What do you think?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
volauvent: No harm to add these two constructors :) For exmaple, I am using `FastSerdeCache(Executor executorService, int limit)` in the test.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeBase.java
volauvent: Sure, will fix it.

PR #84, opened 2020-08-20
Add schema evolution test for UNION field

huangminchn:


FelixGV:
Thanks for the report @huangminchn. This is fixed by #87.

PR #87, opened 2020-08-24
Union overhaul

FelixGV:


avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -530,12 +552,13 @@ private void processUnion(JVar unionSchemaVar, final String name, final Schema u
...
 
-      ifBlock = ifBlock != null ? ifBlock._elseif(condition) : body._if(condition);
-      final JBlock thenBlock = ifBlock._then();
-
       JVar optionSchemaVar = null;
       if (useGenericTypes && unionAction.getShouldRead()) {
-        JInvocation optionSchemaExpression = unionSchemaVar.invoke("getTypes").invoke("get").arg(JExpr.lit(i));
+        if (-1 == readerOptionUnionBranchIndex) {
+          // TODO: Improve the flow of this code... it's messy needing to check for this here...
+          throw new FastSerdeGeneratorException("Unexpected internal state. The readerOptionUnionBranchIndex should be defined if unionAction.getShouldRead() == true.");
```
volauvent, line 559: When this "unexpected internal state" will happen? If so, the following `processComplexType` may also fail due to `Null` `processComplexType`?

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/Array_of_UNION_GenericDeserializer_585074122056792963_585074122056792963.java
```diff
@@ -44,12 +44,15 @@ public Array_of_UNION_GenericDeserializer_585074122056792963_585074122056792963(
...
+                switch (unionIndex0) {
+                    case  0 :
+                        decoder.readNull();
+                        break;
+                    case  1 :
                         array0 .add(deserializerecord0(arrayArrayElementReuseVar0, (decoder)));
-                    }
+                        break;
+                    default:
+                        throw new RuntimeException(("Illegal union index: "+ unionIndex0));
```
radai-rosenblatt, line 55: this exception can be somewhat more useful - print the field name? does decoder have a curreptPosition you could print?

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/Array_of_UNION_GenericDeserializer_585074122056792963_585074122056792963.java
```diff
@@ -67,16 +70,19 @@ public IndexedRecord deserializerecord0(Object reuse, Decoder decoder)
...
-        if (unionIndex1 == 0) {
-            decoder.readNull();
-        } else {
-            if (unionIndex1 == 1) {
+        switch (unionIndex1) {
+            case  0 :
+                decoder.readNull();
+                break;
+            case  1 :
                 if (record.get(0) instanceof Utf8) {
```
radai-rosenblatt, line 78: i dont like this instanceOf Utf8 check, for several reasons:

1. the record could (in theory) be using Strings, or any other Charsequence
2. what this code really wants to do is a null check

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/FastGenericDeserializerGeneratorTest_shouldReadMultipleChoiceUnion_GenericDeserializer_4313387729290221819_4313387729290221819.java
```diff
@@ -40,24 +40,25 @@ public IndexedRecord deserializeFastGenericDeserializerGeneratorTest_shouldReadM
...
-            } else {
-                if (unionIndex0 == 2) {
-                    if (FastGenericDeserializerGeneratorTest_shouldReadMultipleChoiceUnion.get(0) instanceof Utf8) {
-                        FastGenericDeserializerGeneratorTest_shouldReadMultipleChoiceUnion.put(0, (decoder).readString(((Utf8) FastGenericDeserializerGeneratorTest_shouldReadMultipleChoiceUnion.get(0))));
-                    } else {
-                        FastGenericDeserializerGeneratorTest_shouldReadMultipleChoiceUnion.put(0, (decoder).readString(null));
-                    }
+                break;
+            case  2 :
+                if (FastGenericDeserializerGeneratorTest_shouldReadMultipleChoiceUnion.get(0) instanceof Utf8) {
```
radai-rosenblatt, line 51: the getter (get(0)) is called twice. variable?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -492,21 +494,41 @@ private void processUnion(JVar unionSchemaVar, final String name, final Schema u
...
-        readerOptionSchema = readerUnionSchema.getTypes().get(i);
+        // The reader's union could be re-ordered, so we need to find the one that matches.
+        for (int j = 0; j < readerUnionSchema.getTypes().size(); j++) { //  Schema potentialReaderSchema: readerUnionSchema.getTypes()) {
+          Schema potentialReaderSchema = readerUnionSchema.getTypes().get(j);
+          // Avro allows unnamed types to appear only once in a union, but named types may appear multiple times and
+          // thus need to be disambiguated via their full-name (including aliases).
+          if (potentialReaderSchema.getType().equals(optionSchema.getType()) &&
+              (!schemaAssistant.isNamedType(potentialReaderSchema) ||
+                  potentialReaderSchema.getFullName().equals(optionSchema.getFullName()) ||
+                  potentialReaderSchema.getAliases().contains(optionSchema.getFullName()))) {
```
radai-rosenblatt, line 520: are aliases FQCNs or just simple names?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -492,21 +494,41 @@ private void processUnion(JVar unionSchemaVar, final String name, final Schema u
...
+            readerOptionSchema = potentialReaderSchema;
+            readerOptionUnionBranchIndex = j;
```
radai-rosenblatt, line 522: is it worth completing the loop to make sure the result is unambiguous ?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -492,21 +494,41 @@ private void processUnion(JVar unionSchemaVar, final String name, final Schema u
...
+            break;
+          }
+        }
+        if (null == readerOptionSchema) {
+          // This is the same exception that vanilla Avro would throw in this circumstance
+          caseBody._throw(JExpr._new(codeModel.ref(AvroTypeException.class)).arg(
```
radai-rosenblatt, line 528: can you include the field name being handled?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/SchemaAssistant.java
```diff
@@ -137,7 +137,7 @@ public static String getSchemaFullName(Schema schema) {
     return isNamedType ? schema.getFullName() : type.name();
   }
 
-  public static boolean isNamedType(Schema schema) {
+  public static boolean isNamedTypeWithSchema(Schema schema) {
```
radai-rosenblatt, line 140: needs javadoc

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
FelixGV: According to the current logic, the `unionAction.getShouldRead()` condition should be false if `action.getShouldRead()` was also false. If `action.getShouldRead()` is false, then `readerOptionUnionBranchIndex` would still be at -1, but that is fine since `unionAction.getShouldRead()` would prevent us from reaching here...

Furthermore, if we fail to set `readerOptionUnionBranchIndex` then we would also leave `readerOptionSchema == null`, which should likewise make use short-circuit the code here. So I think the current logic is correct, although as I mentioned in the TODO, the flow is quite convoluted and could lead to fragility when future maintenance happens...

I would like to refactor the flow to make things simpler and more robust, but given the time pressure to fix the union bugs, I'd like to minimize the scope of the change in the short-term.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
FelixGV: They are full names, from what I've seen in the Avro code.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
FelixGV: IDK... I guess it should be illegal to have multiple union branches with the same full name. We could construct a test that checks what is the Avro behavior in such cases and emulate the same behavior.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
FelixGV: I would like to have the same exception that Avro throws, to maximize the consistency between vanilla Avro and fast-avro. LMK if you think we should depart from the exceptions thrown by vanilla Avro.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/SchemaAssistant.java
FelixGV: Added. Will be there in the next push.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -492,21 +494,41 @@ private void processUnion(JVar unionSchemaVar, final String name, final Schema u
...
-        ifBlock._then().directStatement(DECODER + ".readNull();");
+        caseBody.directStatement(DECODER + ".readNull();");
+        caseBody._break();
         continue;
       }
 
       if (action.getShouldRead()) {
-        readerOptionSchema = readerUnionSchema.getTypes().get(i);
+        // The reader's union could be re-ordered, so we need to find the one that matches.
+        for (int j = 0; j < readerUnionSchema.getTypes().size(); j++) { //  Schema potentialReaderSchema: readerUnionSchema.getTypes()) {
```
gaojieliu, line 513: Shall we remove these commented out code:
```
//  Schema potentialReaderSchema: readerUnionSchema.getTypes()) {
```

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/Array_of_UNION_GenericDeserializer_585074122056792963_585074122056792963.java
FelixGV: Actually, it is important to check specifically for `Utf8` here because it leads to passing the instance into the `Decoder#readString(Utf8 old)` function. If the old instance is not `Utf8` then we pass null, since other types of `CharSequence` are not mutable and can therefore not be reused.

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/FastGenericDeserializerGeneratorTest_shouldReadMultipleChoiceUnion_GenericDeserializer_4313387729290221819_4313387729290221819.java
FelixGV: I can try... let's see if I manage to do it quickly...

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/Array_of_UNION_GenericDeserializer_585074122056792963_585074122056792963.java
FelixGV: Potentially difficult, but I'll try.

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/Array_of_UNION_GenericDeserializer_585074122056792963_585074122056792963.java
radai-rosenblatt: oh. can code gen add javadocs? :-) i missed the whole reuse part of this

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/FastGenericDeserializerGeneratorTest_shouldReadMultipleChoiceUnion_GenericDeserializer_4313387729290221819_4313387729290221819.java
FelixGV: Ok, I did it. It's non-trivial because in some cases the reuse is already a variable or param, rather than a function invocation, so I only declare new variables where there is an invocation, not when it's already a variable or param. Will be included in the next push.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -492,21 +494,41 @@ private void processUnion(JVar unionSchemaVar, final String name, final Schema u
...
       Schema readerOptionSchema = null;
+      int readerOptionUnionBranchIndex = -1;
       FieldAction unionAction;
-      JExpression condition = unionIndex.eq(JExpr.lit(i));
+      JBlock caseBody = switchBlock._case(JExpr.lit(i)).body();
 
       if (Schema.Type.NULL.equals(optionSchema.getType())) {
-        ifBlock = ifBlock != null ? ifBlock._elseif(condition) : body._if(condition);
-        ifBlock._then().directStatement(DECODER + ".readNull();");
+        caseBody.directStatement(DECODER + ".readNull();");
```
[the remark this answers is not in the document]
FelixGV: We cannot assume that the reader and writer will contain the same number of branches. I don't think we assume that the reader must contain null... here we are enumerating on the writer schema's union branches and providing a case (previously an if) for each branch. This code is transformed from if/else to switch/case but is logically equivalent, it will just invoke `decoder.readNull()` (which is a no-op, BTW).

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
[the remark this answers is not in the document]
FelixGV: I have a test which confirms that vanilla Avro is indeed resilient to union re-ordering, and this PR makes fast-avro consistent with that behavior.

I think what happens in the code snippet you posted is that `in.readIndex()` is a call to the `ResolvingDecoder#readIndex()` which performs a translation of the writer (actual) to reader (expected) index.

FelixGV: Oh yes, thanks for catching that.

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/Array_of_UNION_GenericDeserializer_585074122056792963_585074122056792963.java
FelixGV: Done.

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/Array_of_UNION_GenericDeserializer_585074122056792963_585074122056792963.java
FelixGV: Good question, I don't know :D ...

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/Array_of_UNION_GenericDeserializer_585074122056792963_585074122056792963.java
```diff
@@ -44,12 +44,15 @@ public Array_of_UNION_GenericDeserializer_585074122056792963_585074122056792963(
...
+                switch (unionIndex0) {
+                    case  0 :
+                        decoder.readNull();
+                        break;
+                    case  1 :
                         array0 .add(deserializerecord0(arrayArrayElementReuseVar0, (decoder)));
-                    }
+                        break;
+                    default:
+                        throw new RuntimeException(("Illegal union index for 'arrayElem': "+ unionIndex0));
```
FelixGV, line 55: @radai-rosenblatt this is what you asked for to make the exception more descriptive...

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/Array_of_UNION_GenericDeserializer_585074122056792963_585074122056792963.java
```diff
@@ -67,16 +70,22 @@ public IndexedRecord deserializerecord0(Object reuse, Decoder decoder)
...
-                    record.put(0, (decoder).readString(((Utf8) record.get(0))));
+        switch (unionIndex1) {
+            case  0 :
+                decoder.readNull();
+                break;
+            case  1 :
+            {
+                Object oldString0 = record.get(0);
+                if (oldString0 instanceof Utf8) {
+                    record.put(0, (decoder).readString(((Utf8) oldString0)));
```
FelixGV, line 81: @radai-rosenblatt this is another thing you asked for so that the get invocation happens only once, rather than twice...

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
radai-rosenblatt: i dont think avro is known for having useful exceptions. i think we can retain the same general structure and keywords while providing more info

avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_4/FastGenericDeserializerGeneratorTest_shouldReadEnum_GenericDeserializer_7716892313569022782_7716892313569022782.java
```diff
@@ -46,17 +46,21 @@ public IndexedRecord deserializeFastGenericDeserializerGeneratorTest_shouldReadE
...
+            default:
+                throw new RuntimeException(("Illegal union index for 'testEnumUnion': "+ unionIndex0));
         }
         List<org.apache.avro.generic.GenericData.EnumSymbol> testEnumArray1 = null;
         long chunkLen0 = (decoder.readArrayStart());
-        if (FastGenericDeserializerGeneratorTest_shouldReadEnum.get(2) instanceof List) {
-            testEnumArray1 = ((List) FastGenericDeserializerGeneratorTest_shouldReadEnum.get(2));
+        Object oldArray0 = FastGenericDeserializerGeneratorTest_shouldReadEnum.get(2);
+        if (oldArray0 instanceof List) {
+            testEnumArray1 = ((List) oldArray0);
```
FelixGV, line 63: Note: the invocation caching happens for arrays, maps, strings and fixed. cc @radai-rosenblatt

PR #95, opened 2020-09-22
Fix for ClassNotFoundException occurring during deserializer compilation classpath inference for specific records.

flowenol:
The known type names held by SchemaAssistant.fullyQualifiedClassNameSet also included names for types that had been removed within reader schema. This led to ClassNotFoundException in com.linkedin.avro.fastserde.Utils#inferCompileDependencies method during generation of deserializer for specific record.
This fix adds an additional flag to SchemaAssistant type resolution methods, which allow to omit the inclusion of type name depending on the context.

### DIFF
--- a/avro-fastserde/src/test/java/com/linkedin/avro/fastserde/FastSpecificDeserializerGeneratorTest.java
+++ b/avro-fastserde/src/test/java/com/linkedin/avro/fastserde/FastSpecificDeserializerGeneratorTest.java
@@ -11,8 +11,6 @@ import java.io.IOException;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.nio.ByteBuffer;
-import java.nio.file.Files;
-import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -82,8 +80,7 @@ public class FastSpecificDeserializerGeneratorTest {
 
   @BeforeTest(groups = {"deserializationTest"})
   public void prepare() throws Exception {
-    Path tempPath = Files.createTempDirectory("generated");
-    tempDir = tempPath.toFile();
+    tempDir = getCodeGenDirectory();
 
     classLoader = URLClassLoader.newInstance(new URL[]{tempDir.toURI().toURL()},
         FastSpecificDeserializerGeneratorTest.class.getClassLoader());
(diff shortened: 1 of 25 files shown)


flowenol:
0494221 contains the test case demonstrating the issue
dbe511a contains the fix

avro-fastserde/src/test/java/com/linkedin/avro/fastserde/FastSpecificDeserializerGeneratorTest.java
```diff
@@ -30,6 +31,8 @@
 
 import static com.linkedin.avro.fastserde.FastSerdeTestsSupport.*;
 
+import sun.net.www.content.text.Generic;
```
FelixGV, line 34: This import was added accidentally, right?

FelixGV (approved):
Thanks for contributing this fix and corresponding test!

avro-fastserde/src/test/java/com/linkedin/avro/fastserde/FastSpecificDeserializerGeneratorTest.java
flowenol: Yes, sorry, I must have overlooked that. I'll fix that promptly.

gaojieliu:
@flowenol 
The code change looks good overall.
And I have a recommendation:
Right now, the logic in FastDeserializerGenerator is inconsistent, it doesn't invoke `SchemaAssistant.classFromSchema` in `processRecord` when the incoming field is not required in reader schema, but it does in other type processing, which is causing issue.
Can we invoke the function: `SchemaAssistant.classFromSchema`  only if `action.getShouldRead` is true? I think it will be more consistent.

BTW, can you also commit the generated deserializers for the new test? I think it is very useful to understand the behavior.

flowenol:
@gaojieliu 

Following your suggestion I have modified the fix, so it is much more concise now. Indeed the `classFromSchema` methods should not be called outside `action.getShouldRead() == true` context. In case of `FastDeserializerGenerator` the `processArray` method was the culprit - please review if I haven't included more things than needed in the `action.getShouldRead()` block. Regarding codegen, no classes are generated for tests in `FastSpecificSerializerGeneratorTest`, am I missing something?

gaojieliu:
@flowenol 
Thanks for accommodating the suggestion, the code change looks good to me, and for the test class generation.
If you modify 
```
Path tempPath = Files.createTempDirectory("generated");
    tempDir = tempPath.toFile();
```
in function: `FastSpecificDeserializerGeneratorTest#prepare` to
```
tempDir = getCodeGenDirectory(); 
```
The generated classes will show up in the codegen folder, so we will have a better visibility of the newly changed logic.

gaojieliu (approved):

gaojieliu:
@flowenol 
Please try the latest release:
https://bintray.com/linkedin/maven/avro-util/0.2.10

PR #97, opened 2020-10-04
add serialized payload generation, demonstrate avro json format union incompatibility

radai-rosenblatt:
codegen-X modules in helper tests now can generate not just specific record classes from schemas, but also payloads (serialized records) of those generated records in json and binary wire formats.

added tests that show that avro 1.4 and 1.5 disagree on json format while all avro can read binaries produced by 1.4 and 1.5 - for a single simple schema with an inner record union


FelixGV (approved):
LGTM. Thanks!

PR #101, opened 2020-10-08
Fast-avro support for 1.5, 1.6, 1.9 and 1.10

FelixGV:


helper/impls/helper-impl-16/src/main/java/com/linkedin/avroutil1/compatibility/avro16/Avro16Adapter.java
```diff
@@ -203,6 +203,11 @@ public Object getGenericDefaultValue(Schema.Field field) {
     return Avro16DefaultValuesCache.getDefaultValue(field, false);
   }
 
+  @Override
+  public boolean fieldHasDefault(Schema.Field field) {
+    return null != field.defaultValue();
```
radai-rosenblatt, line 208: what if the default value is null?

avro-fastserde/src/test/java/com/linkedin/avro/fastserde/FastGenericDeserializerGeneratorTest.java
```diff
@@ -468,7 +470,16 @@ public void shouldSkipRemovedField(Implementation implementation) {
 
     // then
     Assert.assertEquals(new Utf8("abc"), record.get("testNotRemoved"));
-    Assert.assertNull(record.get("testRemoved"));
+    if (AvroCompatibilityHelper.getRuntimeAvroVersion().laterThan(AvroVersion.AVRO_1_9)) {
+      try {
+        record.get("testRemoved");
```
radai-rosenblatt, line 475: this relies on avro 1.10+ behavious that throws for fields that dont exist. would it be possibel to have a utility "assertFieldDefined" method in test util ?

helper/impls/helper-impl-16/src/main/java/com/linkedin/avroutil1/compatibility/avro16/Avro16Adapter.java
FelixGV: Then the `field.defaultValue` should be a `NullNode`, and thus not `null`.

avro-fastserde/src/test/java/com/linkedin/avro/fastserde/FastGenericDeserializerGeneratorTest.java
FelixGV: I'm not sure I get the intent of your comment.

Are you saying you'd like this same functionality moved to a utility function (but still be implemented via a try-catch)?

Or are you saying you'd like some other way of detecting if a field is missing (i.e. via reflection or by inspecting the record's schema)?

radai-rosenblatt: option 2. i'd like some Helper.fieldExists(Schema, str) method

avro-fastserde/src/test/java/com/linkedin/avro/fastserde/FastSerdeTestsSupport.java
```diff
@@ -186,4 +190,90 @@ public static File getCodeGenDirectory() throws IOException {
     }
     return dir;
   }
+
+  public static void setField(SpecificRecord sr, String fieldName, Object value) {
+    Class srClass = sr.getClass();
+    try {
+      Field field = srClass.getField(fieldName);
```
radai-rosenblatt, line 197: what happens if the field is private?

radai-rosenblatt (approved):
+1 with minor nits

avro-fastserde/src/test/java/com/linkedin/avro/fastserde/FastSerdeTestsSupport.java
FelixGV: It throws the NoSuchFieldException and proceeds to trying to find the setter function instead.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -37,15 +38,15 @@
...
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericArray;
+import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericFixed;
+import org.apache.avro.generic.GenericRecord;
 import org.apache.avro.generic.IndexedRecord;
 import org.apache.avro.io.Decoder;
-import org.apache.avro.io.parsing.ResolvingGrammarGenerator;
-import org.apache.avro.io.parsing.Symbol;
+import com.linkedin.avro.fastserde.backport.ResolvingGrammarGenerator;
```
gaojieliu, line 46: Having a quick question:
Since for all Avro versions, we are going  to  use the backported classes from  Avro-1.8, so it is might be too advanced for the lower versions than 1.8, and maybe not good enough for higher avro versions than 1.8.
For example, some feature/symbol maybe ignored in vanilla avro-1.4, but could be supported well in the back ported parser/resolver, which will cause the inconsistency between fast-avro and regular avro.

I don't have  any concrete examples here, just think loudly.
Did you observe any inconsistency between fast-avro and vanilla avro after adopting the back ported parser/resolvers?

gaojieliu (approved):
Felix,
Great change, and finally fast-avro supports all the available versions!

Just left some minor comments.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/Utils.java
```diff
@@ -33,6 +33,14 @@ public static boolean isAvro14() {
     return AvroCompatibilityHelper.getRuntimeAvroVersion().equals(AvroVersion.AVRO_1_4);
   }
 
+  public static boolean isAbleToSupportAlternativeStrings() {
+    return AvroCompatibilityHelper.getRuntimeAvroVersion().laterThan(AvroVersion.AVRO_1_6);
+  }
+
+  public static boolean isAbleToSupportJavaStrings() {
```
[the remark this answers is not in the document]
FelixGV: I believe Avro 1.6 added support for Java strings (as opposed to Utf8) in 1.6, and that this applied to all strings (fields, map keys...), then, in 1.7, they further extended this support to other types (URL, URI, etc.). That's why there's a method that checks for > 1.5 and another for >1.6.

I can rename them but I'm not sure what would be a better name. "Stringable prop" does not sound very intuitive either, to me...

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/Utils.java
```diff
@@ -33,6 +33,14 @@ public static boolean isAvro14() {
     return AvroCompatibilityHelper.getRuntimeAvroVersion().equals(AvroVersion.AVRO_1_4);
   }
 
+  public static boolean isAbleToSupportAlternativeStrings() {
```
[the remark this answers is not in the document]
FelixGV: See my response to the other comment below. They are related...

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/Utils.java
```diff
@@ -33,6 +33,14 @@ public static boolean isAvro14() {
...
+    return AvroCompatibilityHelper.getRuntimeAvroVersion().laterThan(AvroVersion.AVRO_1_6);
+  }
+
+  public static boolean isAbleToSupportJavaStrings() {
+    return AvroCompatibilityHelper.getRuntimeAvroVersion().laterThan(AvroVersion.AVRO_1_5);
+  }
+
   public static boolean isSupportedAvroVersionsForDeserializer() {
```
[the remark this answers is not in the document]
FelixGV: Yes, we should add that, and also add support or union defaults... but that is not included in this PR. You made me notice, however, that I had not even added the newly supported versions to the static collections of supported versions...!

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
FelixGV: Well, the reason why I backported it is that there were indeed some significant differences, particularly between 1.8 and 1.9. This part of the Avro code was significantly refactored, allegedly for performance reasons, and it totally broke our code-gen. My understanding is that as long as the binary format is compatible, then these classes should all work interchangeably. I could be wrong of course... not sure.

avro-fastserde/src/test/java/com/linkedin/avro/fastserde/FastGenericDeserializerGeneratorTest.java
FelixGV: @radai-rosenblatt what's the goal of having this new utility function, which would be used purely by the test code?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/Utils.java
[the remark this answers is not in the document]
gaojieliu: I think there  is some disconnect here.
There are  two types of String related support in Avro.
`avro.java.string` was introduced earlier and as you mentioned that it can applied to  all the string fields to specify the right Java type during deserialization, and it is supported in both GenericDatumReader and SpecificDatumReader.
Another string type related feature is called `Stringable Prop`, which is more advanced, since you could not just specify more Java types than `String` for `string` fields, and it is only applicable for SpecificDatumReader, which means these props will be ignored while using GenericDatumReader, and inside Vanilla Avro, it is being called `Stringable Prop`.
That is the original reason I want to clarify which function is for which feature.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/Utils.java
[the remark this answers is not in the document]
gaojieliu: 👍 
Since the PR is already very big, we could create a github issue to track these new features introduced by modern Avro versions.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
gaojieliu: ResolvingGrammarGenerator is mostly used to handle schema parsing/evolution (reader/writer schema), right?
For binary format consistency, I don't doubt it, but more about the schema parsing logic. 
Avro-1.8 could introduce some features, which is not available in Avro-1.4, which means Avro-1.4 ResolvingGrammarGenerator may skip/ignore some symbols, which might  be  added in modern Avro versions, such as in 1.7/1.8. 
Essentially, I am wondering about the inconsistency between Vanilla Avro ResolvingGrammarGenerator and the back ported ResolvingGrammarGenerator from Avro-1.8 when customers are using other Avro versions than 1.8.
Again, I don't have any proof, and it should be fine since all tests pass.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/Utils.java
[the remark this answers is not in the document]
FelixGV: Ok I hadn't realized this was how it was called in vanilla Avro. Then it SGTM to rename `isAbleToSupportAlternativeStrings` to `isAbleToSupportStringableProp` as you suggested.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/Utils.java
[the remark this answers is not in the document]
FelixGV: Yeah, sounds good.

PR #120, opened 2021-01-05
Add FieldBuilder interface and related implementations 

guawang:
1. Add FieldBuilder interface and related implementations in impls 1,4 - 1.9.
2. Add FIELD_BUILDER instance and  one method in AvroCompatibilityHelper to create a Schema.Field by cloning from existing Field.

### DIFF
--- a/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/FieldBuilder.java
+++ b/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/FieldBuilder.java
@@ -17,15 +17,13 @@ import org.apache.avro.Schema.Field.Order;
  */
 public interface FieldBuilder {
 
-  FieldBuilder setField(Schema.Field field);
-
   FieldBuilder setSchema(Schema schema);
 
   FieldBuilder setDoc(String doc);
 
   FieldBuilder setOrder(Order order);
 
-  FieldBuilder copyFromField(Schema.Field field);
+  FieldBuilder copyFromField();
 
   Schema.Field build();
 }
--- a/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/FieldBuilder.java
+++ b/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/FieldBuilder.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2020 LinkedIn Corp.
+ * Licensed under the BSD 2-Clause License (the "License").
+ * See License in the project root for license information.
+ */
+
+package com.linkedin.avroutil1.compatibility;
+
+import org.apache.avro.Schema;
+import org.apache.avro.Schema.Field.Order;
+
+
+/**
+ * an builder interface to build a Schema.Field with given parameters.
+ * The defaultValue needed to be defined in implemented classes since default
+ * value type is changed to Object since AVRO 1.9
+ */
+public interface FieldBuilder {
+
+  FieldBuilder setField(Schema.Field field);
+
+  FieldBuilder setSchema(Schema schema);
+
+  FieldBuilder setDoc(String doc);
+
+  FieldBuilder setOrder(Order order);
+
+  FieldBuilder copyFromField(Schema.Field field);
+
+  Schema.Field build();
+}
--- a/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
+++ b/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
@@ -7,19 +7,12 @@
 package com.linkedin.avroutil1.compatibility;
 
 import com.linkedin.avroutil1.compatibility.avro110.Avro110Adapter;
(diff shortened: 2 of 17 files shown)


helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
```diff
@@ -246,12 +260,7 @@ public static JsonDecoder newJsonDecoder(Schema schema, String in) throws IOExce
 
   /**
    * constructs a {@link JsonDecoder} on top of the given {@link InputStream} for the given {@link Schema}
-   * that is more widely compatible than the "native" avro decoder:
```
radai-rosenblatt, line 249: why strip these docs out ?

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
```diff
@@ -27,20 +39,14 @@
...
-import com.linkedin.avroutil1.compatibility.avro19.Avro19Adapter;
-
 
 /**
  * a Utility class for performing various avro-related operations under a wide range of avro versions at runtime.
  */
 public class AvroCompatibilityHelper {
   private static final AvroVersion DETECTED_VERSION;
   private static final AvroAdapter ADAPTER;
+  private static final FieldBuilder FIELD_BUILDER;
```
radai-rosenblatt, line 49: why is this required? why not add a method - cloneField() into the adapter interface that returns a field builder?

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
```diff
@@ -453,6 +462,18 @@ public static Object getGenericDefaultValue(Schema.Field field) {
...
   }
 
+  /**
+   * returns a new schema field, cloned from an existing schema field and a given schema.
+   * @param field a schema field
+   * @param schema a schema
+   * @return a new field that cloned from the given field and schema
+   * @throws org.apache.avro.AvroRuntimeException if the field in question has no default.
+   */
+  public static Schema.Field cloneSchemaField(Schema.Field field, Schema schema) {
```
radai-rosenblatt, line 472: i think a cleaner API would be:
`public static FieldBuilder cloneField(Schema.Field original)`
users get a builder, and can override any properties they like and call build()

helper/impls/helper-impl-17/src/main/java/com/linkedin/avroutil1/compatibility/avro17/FieldBuilder17.java
```diff
@@ -0,0 +1,64 @@
...
+package com.linkedin.avroutil1.compatibility.avro17;
+
+import com.linkedin.avroutil1.compatibility.FieldBuilder;
+import org.apache.avro.Schema;
+import org.apache.avro.Schema.Field.Order;
+import org.codehaus.jackson.JsonNode;
+
+
+public class FieldBuilder17 implements FieldBuilder {
+  private static Schema.Field _field;
```
radai-rosenblatt, line 16: why is this field static? also i think the Schema.Field (original) should be final and set at constructor time for the builder

guawang: yes it should be final but not static.

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
guawang: I thought for avro-utils, we should not return the "internal" class but return a type in AVRO libraries, like Schema, or Decoder etc.

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
guawang: I thought for avro-utils, we should not return the "internal" class but return a type in AVRO libraries, like Schema, or Decoder etc.

helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/FieldBuilder.java
```diff
@@ -0,0 +1,29 @@
...
+package com.linkedin.avroutil1.compatibility;
+
+import org.apache.avro.Schema;
+import org.apache.avro.Schema.Field.Order;
+
+
+/**
+ * an builder interface to build a Schema.Field with given parameters.
+ * The defaultValue needed to be defined in implemented classes since default
+ * value type is changed to Object since AVRO 1.9
```
radai-rosenblatt, line 16: we could support getting/setting the default value as either Object (primitive-wrapper collection or GenericData.Record), and/or as a json string (cat least as a todo)

radai-rosenblatt (approved):

PR #126, opened 2021-01-12
Always construct deserialized lists with an initial capacity

kcepmada:
Previously this was only done when using generic types.
The initial capacity specified is the size of the first block of an array


radai-rosenblatt (approved):

FelixGV (approved):
LGTM, thanks!

PR #127, opened 2021-01-13
Construct deserialized maps with an initial capacity

kcepmada:
The initial capacity specified is the size of the first block of map entries.


radai-rosenblatt (approved):

FelixGV (approved):
LGTM!

PR #129, opened 2021-01-15
Do not allocate primitive float array unless needed

majisourav99:
Changes in this 

1. Since most of the time avro array format has one block, initialize CompositeByteBuffer to ByteBuffer array of length 1.
2. Since we do not use/read from primitive float[] unless there is a modify operator on the List is called, do not allocate the memory for the `float[]`  it unless modify operator is called.


gaojieliu (approved):
LGTM, thanks Sourav!

PR #139, opened 2021-04-22
Add Schema.Field default value compatibility support in avro 1.4-1.10

SixPlusZero:
Before avro 1.9, the default value in **Schema.Field** constructor takes in **JsonNode** object. Since avro 1.9, it starts taking **Object** as its default value. 
This PR adds a helper function **createSchemaField** in **AvroCompatibilityHelper** to create a Schema.Field instance for all avro versions with the help of FieldBuilder interface.

### DIFF
--- a/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
+++ b/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
@@ -602,4 +602,28 @@ public class AvroCompatibilityHelper {
       return schema.getName();
     }
   }
+
+  /**
+   * Create a new schema field with provided arguments. This adds compatibility layer for {@link Schema.Field} before
+   * avro 1.9, which takes JsonNode as default value in constructor instead of Object.
+   * @param name the name of the field
+   * @param schema the schema of the field
+   * @param doc the doc of the field
+   * @param defaultValue the default value of the field. For avro 1.4-1.8, user should pass in a JsonNode. For avro 1.9
+   * and newer, user should pass in the default value object directly.
+   * @param order the order of the field.
+   * @return fully constructed Field instance
+   */
+  public static Schema.Field createSchemaField(String name, Schema schema, String doc, Object defaultValue, Schema.Field.Order order) {
+    FieldBuilder fieldBuilder = ADAPTER.newFieldBuilder(name);
+    fieldBuilder.setSchema(schema);
+    fieldBuilder.setDoc(doc);
+    fieldBuilder.setDefault(defaultValue);
+    fieldBuilder.setOrder(order);
+    return fieldBuilder.build();
+  }
+
+  public static Schema.Field createSchemaField(String name, Schema schema, String doc, Object defaultValue) {
+    return createSchemaField(name, schema, doc, defaultValue, Schema.Field.Order.ASCENDING);
+  }
 }
--- a/helper/impls/helper-impl-110/src/main/java/com/linkedin/avroutil1/compatibility/avro110/FieldBuilder110.java
+++ b/helper/impls/helper-impl-110/src/main/java/com/linkedin/avroutil1/compatibility/avro110/FieldBuilder110.java
@@ -12,14 +12,20 @@ import org.apache.avro.Schema.Field.Order;
 
 
 public class FieldBuilder110 implements FieldBuilder {
-  private final Schema.Field _field;
+  private final String _name;
+  private Schema.Field _field;
   private Schema _schema;
   private String _doc;
   private Object _defaultVal;
   private Order _order;
 
   public FieldBuilder110(Schema.Field field) {
+    this(field.name());
     _field = field;
+  }
+
+  public FieldBuilder110(String name) {
+    _name = name;
   }
 
   @Override
@@ -34,6 +40,7 @@ public class FieldBuilder110 implements FieldBuilder {
     return this;
   }
 
+  @Override
   public FieldBuilder setDefault(Object defaultValue) {
     _defaultVal = defaultValue;
     return this;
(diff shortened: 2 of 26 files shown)


helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
```diff
@@ -602,4 +609,53 @@ public static String getSchemaFullName(Schema schema) {
...
+   * @param schema the schema of the field
+   * @param doc the doc of the field
+   * @param defaultValue the default value of the field. For avro 1.4-1.8, user should pass in a JsonNode. For avro 1.9
+   * and newer, user should pass in the default value object directly.
+   * @param order the order of the field.
+   * @return fully constructed Field instance
+   */
+  public static Schema.Field createSchemaField(String name, Schema schema, String doc, Object defaultValue, Schema.Field.Order order) {
+    FieldBuilder fieldBuilder;
+    switch (DETECTED_VERSION) {
```
FelixGV, line 626: I think this kind of switching logic belongs inside each adapter.

FelixGV (changes requested):
Thanks @SixPlusZero !

Please take a look at the comment. Also, could we add a test for this?

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
radai-rosenblatt: +1. the general pattern here is delegate all calls to the correct adapter.

helper/impls/helper-impl-110/src/main/java/com/linkedin/avroutil1/compatibility/avro110/FieldBuilder110.java
```diff
@@ -12,16 +12,22 @@
...
 public class FieldBuilder110 implements FieldBuilder {
-  private final Schema.Field _field;
+  private final String _name;
+  private Schema.Field _field;
   private Schema _schema;
   private String _doc;
   private Object _defaultVal;
   private Order _order;
 
   public FieldBuilder110(Schema.Field field) {
```
radai-rosenblatt, line 22: i know this isnt your diff, but can we call this argument "other" ?

helper/impls/helper-impl-14/src/main/java/com/linkedin/avroutil1/compatibility/avro14/FieldBuilder14.java
```diff
@@ -35,6 +41,14 @@ public FieldBuilder setDoc(String doc) {
     return this;
   }
 
+  @Override
+  public FieldBuilder setDefault(Object defaultValue) {
+    if (!(defaultValue instanceof JsonNode)) {
+      throw new RuntimeException("Default value in avro 1.4 should be a JsonNode!");
```
radai-rosenblatt, line 47: 1 - cover nulls
2 - in the exception also print what this thing _IS_ - not just that it isnt a JsonNode.  also - use FQCNs - both jackson1 and 2 have JsonNode and users will be confused.

gaojieliu:
@radai-rosenblatt 
I saw there are a lot of duplications among FieldBuilder14 ... FieldBuilder10, so any specific reason we couldn't abstract the common codes/functions?

helper/impls/helper-impl-110/src/main/java/com/linkedin/avroutil1/compatibility/avro110/FieldBuilder110.java
SixPlusZero: Hi @radai-rosenblatt, are you referring to the **_field** to be changed to **_otherField**? If so I could also change it.

helper/impls/helper-impl-14/src/main/java/com/linkedin/avroutil1/compatibility/avro14/FieldBuilder14.java
SixPlusZero: Right. This logic no longer exists in the newest diff, as we take in real default value object and perform object -> JsonNode conversion now.

FelixGV (approved):
Thanks! LGTM. Let's merge this on Monday if no one else objects.

helper/impls/helper-impl-110/src/main/java/com/linkedin/avroutil1/compatibility/avro110/FieldBuilder110.java
radai-rosenblatt: no, the constructor argument

helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/JsonUtils.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class JsonUtils {
+  static final String BYTES_CHARSET = "ISO-8859-1";
+
+  /**
+   * toJsonNode takes in an arbitrary object and convert it into corresponding JsonNode
+   * @param datum input datum object
+   * @return generated JsonNode object.
+   */
+  public static JsonNode toJsonNode(Object datum) {
```
radai-rosenblatt, line 26: this method should probably have Jackson1 in the name, since it returns jackson 1 objects (jackson 2 is under the fasterxml package)

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
```diff
@@ -602,4 +602,28 @@ public static String getSchemaFullName(Schema schema) {
...
     }
   }
+
+  /**
+   * Create a new schema field with provided arguments. This adds compatibility layer for {@link Schema.Field} before
+   * avro 1.9, which takes JsonNode as default value in constructor instead of Object.
+   * @param name the name of the field
+   * @param schema the schema of the field
+   * @param doc the doc of the field
+   * @param defaultValue the default value of the field. For avro 1.4-1.8, user should pass in a JsonNode. For avro 1.9
```
radai-rosenblatt, line 612: you should clarify which methods on avro class Schema() or elsewhere these defauult valuues should come from

radai-rosenblatt (approved):

Issue #151, opened 2021-06-07
when fixing up code generated by 1.9+, remove ref. to AvroMissingFieldException from builder build() methods


radai-rosenblatt:
code generated under avro 1.9+ has a catch clause for AvroMissingFieldException in Builder.build() methods.
this exception class does not exist prior to 1.9 so generated code causes issues when run under avro < 1.9.

solution is to strip out the catch clause.

kxu1026:
Thanks, Radai!

PR #152, opened 2021-06-07
remove references to new avr exception classes from builders for broader compatibility

radai-rosenblatt:
addresses #151

this code replaces this bit:
```java
catch (org.apache.avro.AvroMissingFieldException e) {
   throw e;
} catch (java.lang.Exception e) {
   throw new org.apache.avro.AvroRuntimeException(e);
}
```
which references AvroMissingFieldException - that only exists in avro 1.9+ - with this:
```java
catch (java.lang.Exception e) {
   throw e instanceof org.apache.avro.AvroRuntimeException ? (org.apache.avro.AvroRuntimeException) e : new org.apache.avro.AvroRuntimeException(e);
}
```

### DIFF
--- a/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/CodeTransformations.java
+++ b/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/CodeTransformations.java
@@ -74,6 +74,11 @@ public class CodeTransformations {
   private static final Pattern IMPORT_PATTERN = Pattern.compile("import\\s+(.*);");
   private static final Pattern ENUM_PATTERN = Pattern.compile("public enum (\\w+)");
   private static final Pattern ANNOTATION_PATTERN = Pattern.compile("\\s*@.*");
+  private static final Pattern CATCH_FIELD_EXCEPTION_START_PATTERN = Pattern.compile("catch \\(org.apache.avro.AvroMissingFieldException e\\) \\{");
+  private static final Pattern CATCH_FIELD_EXCEPTION_END_PATTERN = Pattern.compile("throw new org.apache.avro.AvroRuntimeException\\(e\\);\\s*}");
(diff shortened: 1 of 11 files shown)


helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/CodeTransformations.java
```diff
@@ -74,6 +74,11 @@
   private static final Pattern IMPORT_PATTERN = Pattern.compile("import\\s+(.*);");
   private static final Pattern ENUM_PATTERN = Pattern.compile("public enum (\\w+)");
   private static final Pattern ANNOTATION_PATTERN = Pattern.compile("\\s*@.*");
+  private static final Pattern CATCH_FIELD_EXCEPTION_START_PATTERN = Pattern.compile("catch \\(org.apache.avro.AvroMissingFieldException e\\) \\{");
+  private static final Pattern CATCH_FIELD_EXCEPTION_END_PATTERN = Pattern.compile("throw new org.apache.avro.AvroRuntimeException\\(e\\);\\s*}");
```
navina, line 78: How do we know that there will always be a `throw new AvroRuntimeException`  at the end of the clause list? 
Maybe just remove the catch clause of the `AvroMissingFieldException` ?

navina (approved):
one question. otherwise, +1

helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/CodeTransformations.java
radai-rosenblatt: just removing the top catch clause would mean that AvroMissingFieldExceptions would always be double-wrapped in an AvroRuntimeException, which is a change of behaviour under 1.9/10. 

we know this pattern holds for 1.9 and 1.10. if something new happens in 1.11+ we'll have to adjust. the theoretically correct thing to do is stop mutilating code post-avro and throw avro code0gen out the window entirely and replace it with our own, but thats a longer-term todo

Issue #146, opened 2021-05-24
Maven cannot resolve avroutil1:helper:0.2.32 dependency


narner90:
I get the following error from Maven when including avro-codegen version 0.2.32:
```
Cannot resolve com.linkedin.avroutil1:helper:0.2.32
```

My dependency is specified as: 
```
<dependency>
    <groupId>com.linkedin.avroutil1</groupId>
    <artifactId>avro-codegen</artifactId>
    <version>0.2.32</version>
    <type>pom</type>
</dependency>
```

And the repository as:
```
<repository>
    <id>linkedin</id>
    <name>linkedin</name>
    <url>https://linkedin.bintray.com/maven/</url>
</repository>
```

Here is the relevant snippet from the avro-codegen POM that maven downloads. **Should `helper` be `helper-all` here?**
```
<dependency>
      <groupId>com.linkedin.avroutil1</groupId>
      <artifactId>helper</artifactId>
      <version>0.2.32</version>
      <scope>runtime</scope>
</dependency>
```

Thanks!

radai-rosenblatt:
we do not publish the "thin" artifact.
have you tried helper-all ?

narner90:
Thanks for the reply - `helper-all` works fine on its own for me.

I suspect it is the `avro-codegen` POM [bintray link](https://bintray.com/linkedin/maven/avro-util#files/com%2Flinkedin%2Favroutil1%2Favro-codegen%2F0.2.33) is the issue here, as it specifies `helper` rather than `helper-all` as a dependency.

radai-rosenblatt:
you are correct. the pom produced for codegen (https://linkedin.jfrog.io/artifactory/avro-util/com/linkedin/avroutil1/avro-codegen/0.2.35/avro-codegen-0.2.35.pom) refers to helper and not helper-all.
i'll see what can be done to fix that

radai-rosenblatt:
should be fixed now

PR #155, opened 2021-06-11
fix published dependencies for codegen, publish spotbugs, kill bintray bits

radai-rosenblatt:
should also address #146

### DIFF
--- a/avro-codegen/build.gradle
+++ b/avro-codegen/build.gradle
@@ -50,3 +50,88 @@ shadowJar {
     //needs to have a classifier to not overwrite the regular jar
     classifier = "all"
 }
+
+publishing {
+    publications {
+        maven(MavenPublication) {
+            groupId project.group
+            artifactId project.name
+            version project.version
+
+            from components.java
+            artifact sourceJar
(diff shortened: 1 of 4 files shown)


avro-codegen/build.gradle
```diff
@@ -49,4 +49,89 @@ jar {
...
+                description = 'utilities for writing code that works across major avro versions'
+                url = 'https://github.com/linkedin/avro-util'
+
+                licenses {
+                    license {
+                        name = 'BSD 2-Clause'
+                        url = 'https://raw.githubusercontent.com/linkedin/avro-util/master/LICENSE'
+                    }
+                }
+                developers {
```
FelixGV, line 77: Have these folks ever contributed to the project 🤣 ...?

FelixGV (approved):

avro-codegen/build.gradle
radai-rosenblatt: yes, thayve also contributed to the helper codebase back when it was an internal project and so github history is distorted.
we really need some gradle magic to not repeat this pom blurb so many time though ...

PR #163, opened 2021-06-25
initial SchemaBuilder support

radai-rosenblatt:


FelixGV (commented):
I don't get it... what's the point of the `SchemaBuilder` interface if its only function is `build()`? Are you planning to add more functions later?

radai-rosenblatt:
getters/setters can always be added to the builder when needed (like FieldBuilder).
the functionality in this PR is enough to fully clone a schema under all supported avro versions. beyond that Schemas are actually mutable so some small changes can be made after build()

radai-rosenblatt:
@FelixGV added add/remove field methods

helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/SchemaBuilder.java
```diff
@@ -0,0 +1,22 @@
...
+
+package com.linkedin.avroutil1.compatibility;
+
+import org.apache.avro.Schema;
+
+public interface SchemaBuilder {
+
+    SchemaBuilder addField(Schema.Field field);
+
+    SchemaBuilder addField(int position, Schema.Field field);
```
FelixGV, line 15: This will shift fields to the right, right? We should document the behavior.

helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/SchemaBuilder.java
```diff
@@ -0,0 +1,22 @@
...
+
+    SchemaBuilder removeField(String fieldName);
+
+    SchemaBuilder removeField(int position);
```
FelixGV, line 19: Same here. Actually, some short JavaDoc for all functions would be great.

FelixGV (approved):
LGTM. Please add JavaDoc to the new interface before merging.

PR #171, opened 2021-07-18
start work on parser module - implement a location-preserving json parser

radai-rosenblatt:
this is a building block on the way to implementing our own parser for avsc files.
having a json parser that can trace elements back to their line/column/character is key for better error messages to users.
json-p was chosen as jackson is "tainted" because avro uses different (incompatible) versions of it and we want to coexist with avro (and we dont want to shade).

### DIFF
--- a/helper/tests/helper-tests-16/src/test/java/com/linkedin/avroutil1/compatibility/avro16/AvroCompatibilityHelperAvro16Test.java
+++ b/helper/tests/helper-tests-16/src/test/java/com/linkedin/avroutil1/compatibility/avro16/AvroCompatibilityHelperAvro16Test.java
@@ -7,7 +7,7 @@
 package com.linkedin.avroutil1.compatibility.avro16;
 
 import com.linkedin.avroutil1.Pojo;
-import com.linkedin.avroutil1.TestUtil;
+import com.linkedin.avroutil1.testcommon.TestUtil;
 import com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper;
 import com.linkedin.avroutil1.compatibility.AvroVersion;
 import java.io.IOException;
--- a/parser/src/main/java/com/linkedin/avroutil1/parser/jsonpext/JsonReaderExt.java
+++ b/parser/src/main/java/com/linkedin/avroutil1/parser/jsonpext/JsonReaderExt.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright 2021 LinkedIn Corp.
+ * Licensed under the BSD 2-Clause License (the "License").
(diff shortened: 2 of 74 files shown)


helper/tests/helper-tests-16/src/test/java/com/linkedin/avroutil1/compatibility/avro16/AvroCompatibilityHelperAvro16Test.java
```diff
@@ -7,7 +7,7 @@
 package com.linkedin.avroutil1.compatibility.avro16;
 
 import com.linkedin.avroutil1.Pojo;
-import com.linkedin.avroutil1.TestUtil;
+import com.linkedin.avroutil1.testcommon.TestUtil;
```
navina, line 10: Can you separate out the move of the class to a new package namespace in a separate PR? we can avoid ~40 files in this diff?

parser/src/main/java/com/linkedin/avroutil1/parser/jsonpext/JsonReaderExt.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+package com.linkedin.avroutil1.parser.jsonpext;
+
+import jakarta.json.JsonReader;
+
+public interface JsonReaderExt extends JsonReader {
+
+    @Override
+    JsonStructureExt read();
```
navina, line 14: how is this overriden API different from the one in `jakarta.json.JsonReader` ?
It also looks like `read` and `readArray` is unsupported. Why is that the case?

radai-rosenblatt: returns JsonStructureExt and not JsonStructure. others unimplemented because i didnt need them (yet?)

helper/tests/helper-tests-16/src/test/java/com/linkedin/avroutil1/compatibility/avro16/AvroCompatibilityHelperAvro16Test.java
radai-rosenblatt: will do

parser/src/main/java/com/linkedin/avroutil1/parser/jsonpext/JsonReaderExt.java
navina: Ok. maybe add some doc for this?

radai-rosenblatt:
replaced by https://github.com/linkedin/avro-util/pull/176

PR #182, opened 2021-08-23
Fix the handling of null default values.

srramach:
A null argument for the default value param in `createSchemaField` could
mean one of two things:
(1) The default value was not specified, or
(2) The default value was specified to be null.

To avoid bloating the helper API, we disambiguate between the two cases
based on whether the field schema allows null as a default value.

Prior to this PR, the helper's behaviour was inconsistent. In Avro 1.8
and earlier, we always converted a null argument into a NullNode object.
In Avro 1.9 and later, null was never conveted. This PR brings makes
them behave consistently.

### DIFF
--- a/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/FieldBuilder.java
+++ b/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/FieldBuilder.java
@@ -21,7 +21,7 @@ public interface FieldBuilder {
 
   FieldBuilder setDoc(String doc);
 
-  FieldBuilder setDefault(Object defaultValue, Schema schema);
+  FieldBuilder setDefault(Object defaultValue);
 
   FieldBuilder setOrder(Order order);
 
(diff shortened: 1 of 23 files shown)


helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/FieldBuilder.java
```diff
@@ -21,7 +21,7 @@
 
   FieldBuilder setDoc(String doc);
 
-  FieldBuilder setDefault(Object defaultValue);
+  FieldBuilder setDefault(Object defaultValue, Schema schema);
```
radai-rosenblatt, line 24: the field has a schema. accepting a potentially different schema as arg here opens the door to various bugs where field schema and default value schema are completely unrelated.

maybe instead we assert that the field schema has already been set at this point, and derive the default value schema from there?

srramach: Done. Also added better comments, more tests and updated the PR description. PTAL.

radai-rosenblatt (approved):

PR #194, opened 2021-09-09
Added a flag in FastDatumReader/FastDatumWriter to indicate whether fast class is being used or not

gaojieliu:


gaojieliu:
cc @radai-rosenblatt

radai-rosenblatt (approved):

PR #201, opened 2021-09-24
preserve root cause of SpecificCompiler instantiation failure, also avro 1.10.2

### LINKED ISSUE #197: print _root_ cause exception when avro compiler not found on classpath
sometimes its due to something _else_ (not avro-compiler jar) being excluded from dependencies.

radai-rosenblatt:
resolves #197


FelixGV (approved):
LGTM!

PR #205, opened 2021-10-01
fix detection of exact avro 1.7 version under classpaths with multiple avro jars

### LINKED ISSUE #203: Don't use the existence of one class to infer things about OTHER classes.
In Avro17Utils.java, we check for the existence of `org.apache.avro.JsonProperties` to see if our runtime Avro is v1.7.3 or later. Assuming that's true, we proceed to operate on the `Schema.Field` that we are given, expecting that the Field is a subclass of JsonProperties.

But that can throw an exception if an older version of Avro is earlier on the classpath, and Avro 1.7.3+ is merely present later on the classpath. I.e., Schema.Field comes from Avro 1.7.2 or earlier, and thus does *not* extend JsonProperties. But the class loader will find JsonProperties, since it does exist on the classpath.

Here's an example of the exception that's thrown:
```
Exception in thread "main" java.lang.IllegalStateException: java.lang.IllegalArgumentException: object is not an instance of declaring class
	at com.linkedin.avroutil1.compatibility.avro17.Avro17Utils.getJsonProp(Avro17Utils.java:80)
	at com.linkedin.avroutil1.compatibility.avro17.Avro17Adapter.getFieldPropAsJsonString(Avro17Adapter.java:361)
	at com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper.getFieldPropAsJsonString(AvroCompatibilityHelper.java:728)
	at com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper.getFieldPropAsJsonString(AvroCompatibilityHelper.java:713)
	...
```

This is not an uncommon scenario. In our company's codebase, we do often see classpaths where multiple versions of Avro are present.

There may be other instances of such incorrect determination, elsewhere in the compat helper codebase. I'll update this bug as I find them.

radai-rosenblatt:
resolves #203

### DIFF
--- a/helper/impls/helper-impl-17/src/main/java/com/linkedin/avroutil1/compatibility/avro17/Avro17Utils.java
+++ b/helper/impls/helper-impl-17/src/main/java/com/linkedin/avroutil1/compatibility/avro17/Avro17Utils.java
@@ -36,13 +36,19 @@ public class Avro17Utils {
     }
 
     static boolean isIsAtLeast173() {
-        try {
-            //this class was created as part of AVRO-1157 for 1.7.3
-            Class.forName("org.apache.avro.JsonProperties");
+        //class org.apache.avro.JsonProperties was created as part of AVRO-1157 for 1.7.3
+        //however, if we naively just test for its existence we risk finding it in some extra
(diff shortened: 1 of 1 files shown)


srramach (approved):
LGTM

PR #214, opened 2021-10-03
parse logical types and string representation

radai-rosenblatt:
"negative" test cases will be in a follow-up PR

### DIFF
--- a/parser/src/main/java/com/linkedin/avroutil1/model/AvroPrimitiveSchema.java
+++ b/parser/src/main/java/com/linkedin/avroutil1/model/AvroPrimitiveSchema.java
@@ -11,13 +11,30 @@ package com.linkedin.avroutil1.model;
  */
 public class AvroPrimitiveSchema extends AvroSchema {
     private final AvroType type;
+    private final AvroLogicalType logicalType;
+    private final AvroJavaStringRepresentation javaStringRepresentation; //only valid for string schemas
 
-    public AvroPrimitiveSchema(CodeLocation codeLocation, AvroType type) {
+    public AvroPrimitiveSchema(
+            CodeLocation codeLocation,
+            AvroType type,
+            AvroLogicalType logicalType,
+            AvroJavaStringRepresentation javaStringRepresentation
+    ) {
         super(codeLocation);
         if (!type.isPrimitive()) {
             throw new IllegalArgumentException(type + " is not a primitive type");
         }
+        if (logicalType != null && !logicalType.getParentTypes().contains(type)) {
+            throw new IllegalArgumentException(type() + " at " + codeLocation + " cannot have a logical type of "
+                    + logicalType + " (which can only be a logical type of " + logicalType.getParentTypes() + ")");
+        }
+        if (javaStringRepresentation != null && !AvroType.STRING.equals(type)) {
+            throw new IllegalArgumentException("cant set a javaStringRepresentation (" + javaStringRepresentation
+                    + ") for type " + type + " as it is not a string");
+        }
         this.type = type;
+        this.logicalType = logicalType;
+        this.javaStringRepresentation = javaStringRepresentation;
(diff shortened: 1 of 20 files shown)


parser/src/main/java/com/linkedin/avroutil1/model/AvroPrimitiveSchema.java
```diff
@@ -11,26 +11,74 @@
...
+    ) {
         super(codeLocation);
         if (!type.isPrimitive()) {
             throw new IllegalArgumentException(type + " is not a primitive type");
         }
+        if (logicalType != null && !logicalType.getParentTypes().contains(type)) {
+            throw new IllegalArgumentException(type() + " at " + codeLocation + " cannot have a logical type of "
+                    + logicalType + " (which can only be a logical type of " + logicalType.getParentTypes() + ")");
+        }
+        if (javaStringRepresentation != null && !AvroType.STRING.equals(type)) {
```
srramach, line 31:
```suggestion
        if (javaStringRepresentation != null && type != AvroType.STRING) {
```

parser/src/main/java/com/linkedin/avroutil1/model/AvroPrimitiveSchema.java
```diff
@@ -11,26 +11,74 @@
...
         return type;
     }
 
+    @Override
+    public AvroLogicalType logicalType() {
+        return logicalType;
+    }
+
+    public AvroJavaStringRepresentation getJavaStringRepresentation() {
+        if (!AvroType.STRING.equals(type())) {
```
srramach, line 51:
```suggestion
        if (type != AvroType.STRING) {
```

parser/src/main/java/com/linkedin/avroutil1/model/AvroPrimitiveSchema.java
```diff
@@ -11,26 +11,74 @@
...
 
-    public static AvroPrimitiveSchema forType(CodeLocation codeLocation, AvroType type) {
-        return new AvroPrimitiveSchema(codeLocation, type);
+    @Override
+    public String toString() {
+        if (javaStringRepresentation == null || !AvroType.STRING.equals(type)) {
+            return super.toString();
+        }
+        //reflect avro string type for string schemas
+        String typeName = javaStringRepresentation.getJsonValue();
```
srramach, line 68: Not a good idea. Introduces unnecessary subtleties and differences when reading the debug/error messages.

I.e., in principle, since `CharSequence` is the default, adding a `"avro.java.string": "CharSequence"` property changes nothing about the schema or the generated code, but now all the debug/error messages will say `CharSequence schemaName` instead of `string schemaName`. Likewise, it's too confusing to distinguish, in said debug/error messages, between the default (`string schemaName`) and when the property has been set to String (`String schemaName`).

I think it's better to add another suffix, say "(javaType=<JavaType>)", just like with logicalType. That also means you can avoid reinventing the wheel below (w.r.t. attaching the logicalType info), and just use super.toString() and then attach the string type details, if necessary.

parser/src/main/java/com/linkedin/avroutil1/model/AvroPrimitiveSchema.java
```diff
@@ -11,26 +11,74 @@
...
     @Override
     public boolean isNullable() {
         return type == AvroType.NULL;
     }
 
-    public static AvroPrimitiveSchema forType(CodeLocation codeLocation, AvroType type) {
-        return new AvroPrimitiveSchema(codeLocation, type);
+    @Override
+    public String toString() {
+        if (javaStringRepresentation == null || !AvroType.STRING.equals(type)) {
```
srramach, line 64:
```suggestion
        if (javaStringRepresentation == null || type != AvroType.STRING) {
```

PR #224, opened 2021-10-12
create value object only when cache does not have the key-value pair

yuzheng21:
putIfAbsent will always evaluate the parameter
use computeIfAbsent instead

### DIFF
--- a/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
+++ b/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
@@ -362,7 +362,7 @@ public final class FastSerdeCache {
       LOGGER.warn("Deserializer class instantiation exception", e);
     }
 
-    return new FastDeserializer<>() {
+    return new FastDeserializer<Object>() {
       private DatumReader datumReader = new SpecificDatumReader<>(writerSchema, readerSchema);
 
       @Override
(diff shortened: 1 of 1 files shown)


radai-rosenblatt (approved):

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -206,14 +207,17 @@ public static boolean isSupportedForFastSerializer(Schema.Type schemaType) {
     FastDeserializer<?> deserializer = fastSpecificRecordDeserializersCache.get(schemaKey);
 
     if (deserializer == null) {
-      deserializer = fastSpecificRecordDeserializersCache.putIfAbsent(schemaKey,
-          new FastDeserializerWithAvroSpecificImpl<>(writerSchema, readerSchema));
-      if (deserializer == null) {
-        deserializer = fastSpecificRecordDeserializersCache.get(schemaKey);
+      AtomicBoolean status = new AtomicBoolean(false);
+      deserializer = fastSpecificRecordDeserializersCache.computeIfAbsent(
+          schemaKey,
```
gaojieliu, line 212: This seems too complicate.
Shall we just move the following logic into the `computeIfAbsent`?
```
CompletableFuture.supplyAsync(() -> buildSpecificDeserializer(writerSchema, readerSchema), executor)
            .thenAccept(d -> fastSpecificRecordDeserializersCache.put(schemaKey, d));
```

yuzheng21: `d -> fastSpecificRecordDeserializersCache.put(schemaKey, d)` is inside computeIfAbsent, this update will fail due to nested update not allowed exception.

Issue #222, opened 2021-10-07
Fast-avro should support enum default value with avro-1.9+


gaojieliu:
Right now, Fast-avro is not respecting the enum default value for all avro versions, which is not right.
This issue is used to track the enum default value support for avro-1.9+.

gaojieliu:
cc @yuzheng21

FelixGV:
I was thinking about that as well... maybe we ought to be have the 1.9 behavior across all versions. The only catch is that there is a warm up phase, and until it's complete, we call regular (well, mostly regular) Avro instead... so that means there could be brief periods of time where the pre-1.9 behavior shows up, which is less than ideal.

radai-rosenblatt:
if i had to guess i'd say 99% of users would be fine with blocking until a codec class is ready. maybe we could also change behaviour there (make use-vanilla-until-warmup an opt-in)

PR #225, opened 2021-10-13
add enum default support at fast-serdese

yuzheng21:
add enum default support at fast-serdese

https://github.com/linkedin/avro-util/issues/222

### DIFF
--- a/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/backport/ResolvingGrammarGenerator.java
+++ b/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/backport/ResolvingGrammarGenerator.java
@@ -17,6 +17,7 @@
  */
 package com.linkedin.avro.fastserde.backport;
 
+import com.linkedin.avro.fastserde.Utils;
 import com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
@@ -48,7 +49,7 @@ public class ResolvingGrammarGenerator extends ValidatingGrammarGenerator {
    */
   public final Symbol generate(Schema writer, Schema reader)
     throws IOException {
-    return Symbol.root(generate(writer, reader, new HashMap<LitS, Symbol>()));
+    return Symbol.root(generate(writer, reader, new HashMap<>()));
   }
 
   /**
@@ -56,7 +57,7 @@ public class ResolvingGrammarGenerator extends ValidatingGrammarGenerator {
    * reader and returns the start symbol for the grammar generated.
    * If there is already a symbol in the map seen for resolving the
    * two schemas, then that symbol is returned. Otherwise a new symbol is
-   * generated and returnd.
+   * generated and returned.
    * @param writer    The schema used by the writer
    * @param reader    The schema used by the reader
    * @param seen      The &lt;reader-schema, writer-schema&gt; to symbol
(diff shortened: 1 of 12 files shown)


avro-fastserde/src/main/java/com/linkedin/avro/fastserde/backport/ResolvingGrammarGenerator.java
```diff
@@ -448,6 +451,16 @@ private static Symbol mkEnumAdjust(List<String> wsymbols,
     return Symbol.enumAdjustAction(rsymbols.size(), adjustments);
   }
 
+  private static Symbol mkEnumAdjustWithDefault(Schema writer, Schema reader) {
+    Resolver.EnumAdjust e = (Resolver.EnumAdjust) Resolver.EnumAdjust.resolve(writer, reader, GenericData.get());
```
gaojieliu, line 455: Instead of backporting Resolver from avro-1.9, can we add this method to AvroCompatiblityHelper, which should be already adopting the version-specific logic for each Avro version?

@radai-rosenblatt Could you share your opinion on this as well?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/backport/ResolvingGrammarGenerator.java
```diff
@@ -448,6 +451,16 @@ private static Symbol mkEnumAdjust(List<String> wsymbols,
...
+    Object[] adjs = new Object[e.adjustments.length];
+    for (int i = 0; i < adjs.length; i++) {
+      adjs[i] = (0 <= e.adjustments[i] ? new Integer(e.adjustments[i])
```
gaojieliu, line 458: It is a little weird to write this way, so can you modify it to be `e.adjustments[i] >= 0` instead?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/backport/Symbol.java
```diff
@@ -481,10 +481,20 @@ public static EnumAdjustAction enumAdjustAction(int rsymCount, Object[] adj) {
...
+    public final boolean noAdjustments;
     public final Object[] adjustments;
+
     @Deprecated public EnumAdjustAction(int rsymCount, Object[] adjustments) {
       super(rsymCount);
       this.adjustments = adjustments;
+      boolean noAdj = true;
+      if (adjustments != null) {
+        int count = Math.min(rsymCount, adjustments.length);
+        noAdj = (adjustments.length <= rsymCount);
```
gaojieliu, line 493: Why not use `>=` here to make it more readable?

gaojieliu: Also can you leave some comments on these new logics?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/backport/ResolvingGrammarGenerator.java
yuzheng21: I just copied the logic from Avro

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/backport/Symbol.java
yuzheng21: Same here, I just copied this from avro Symbol class

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/backport/ResolvingGrammarGenerator.java
radai-rosenblatt: this API is used internally at linkedin. im ok with adding a bollean flag to the newCachedResolvingDecoder() method to control support for default enum values (which we could honor under any avro version)

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/backport/ResolvingGrammarGenerator.java
```diff
@@ -448,6 +451,16 @@ private static Symbol mkEnumAdjust(List<String> wsymbols,
...
+          : "No match for " + e.writer.getEnumSymbols().get(i));
```
radai-rosenblatt, line 459: this error msg could be a lot more useful: "reader schema {name} has no match for writer schema enum symbol XXX (value YYY)"

yuzheng21: From my understanding, this is not an error message.
Vanilla Avro sets the same enum symbol for unknown values
"
No match for ${enum value}
"

helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvroAdapter.java
```diff
@@ -114,6 +116,17 @@ default FieldBuilder newFieldBuilder(String name) {
...
 
+  default String getEnumDefault(Schema s) {
+    throw new AvroTypeException("enum default is not supported in " + supportedMajorVersion().toString());
+  }
+
+  default Schema newEnumSchema(String name, String doc, String namespace, List<String> values, String enumDefault) {
+    if (enumDefault != null) {
+      throw new AvroTypeException("enum default is not supported in " + supportedMajorVersion().toString());
+    }
+    return Schema.createEnum(name, doc, namespace, values);
```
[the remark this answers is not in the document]
yuzheng21: here is to check if the enum default feature is supported by the underlying avro schema.
For example, for avro < 1.9, the adapter should throw such exception if the user tries to create a enumSchema with a non-null enumDefault.

PR #227, opened 2021-10-17
add parser support for schema and field properties

radai-rosenblatt:
prop order is preserved (as defined in the source schema).
to avoid any issues with exposing json libraries in the API prop access is as json literals or java pojos (gson-style)


parser/src/main/java/com/linkedin/avroutil1/model/AvroUnionSchema.java
```diff
@@ -21,7 +21,7 @@
     private List<SchemaOrRef> types;
 
     public AvroUnionSchema(CodeLocation codeLocation) {
-        super(codeLocation);
+        super(codeLocation, null); //cant put extra props on a union
```
srramach, line 24: There are no null checks in the base class (AvroSchema) when accessing props-related methods (propertyNames(), getPropertyAsObject(), etc). I'm not sure if those methods can be called under _some_ callpath for unions; but to be safe, maybe pass in an JsonPropertiesContainer.EMPTY instead of null?

parser/src/main/java/com/linkedin/avroutil1/parser/avsc/AvscParser.java
```diff
@@ -246,19 +286,20 @@ private SchemaOrRef parseComplexSchema(
                     + typeStr.getLocation() + ". expecting \"record\", \"enum\" or \"fixed\"");
         }
 
+        LinkedHashMap<String, JsonValueExt> propsMap = parseExtraProps(objectNode, CORE_SCHEMA_PROPERTIES);
+        JsonPropertiesContainerImpl props = new JsonPropertiesContainerImpl(propsMap);
```
srramach, line 290: Why not do the same thing as you do on line 401? I.e. if the props is empty, use JsonPropertiesContainer.EMPTY instead of creating an object of the impl.

srramach (approved):
LGTM with a couple of minor comments.

PR #243, opened 2021-11-10
warm up datum readers

yuzheng21:
warm up datum readers for Generic and Specific Deserializer
- change cachedFastDeserializer to AtomicReference to avoid race condition


radai-rosenblatt:
whats the purpose of warmup ?
how can a user tell if warmup completed within timeout or not?

radai-rosenblatt:
i seriously dislike this API extention.
the correct solution in my opinion isto have all getFastWhatever() methods return a future, which users could block on as "warm up".
for users of the current "sneaky" fast codecs its possible to write a delegate backed by the above future and a stop-gap vanilla avro codec.

allowing this PR in instead of a proper API would only make any future refactor worse

yuzheng21:
Thank you @radai-rosenblatt for your feedback. I agree with you.
It looks like there are two separate issues
1. getFastSpecificDeserializer - apply delegate pattern to return Future
2. refactor current behavior - trigger build until the user calls read

for 1 , that can support warm up function at user side as you mentioned
for 2 , that may have some performance impact at user side since it changes currently lazy build behavior

Thus, shall we track these two things separately?

radai-rosenblatt:
i think the current behaviour of only starting work on the codec on read() is horrible, and likely the reason you started implementing this warmup code in the 1st place.

work on the codec class should start when its created.

gaojieliu:
> i think the current behaviour of only starting work on the codec on read() is horrible, and likely the reason you started implementing this warmup code in the 1st place.
> 
> work on the codec class should start when its created.

@radai-rosenblatt 
The original purpose of this behavior is to fall back to Vanilla Avro when the fast-class is not available yet.
In a busy node, the fast class generation could take several seconds, and we don't want to delay the user's request util the codec generation is done.
For the cases, who know all the schema pairs, it might be reasonable to generate the codec during startup, but for other cases, where not all the schema pairs are known ahead, we don't want block the user's request when generating the fast codec at runtime.
So I think we could do two things:
1. Move the async codec gen from the read function to the constructor function.
2. Introduce a new param in fast reader/writer to decide whether the fast codec generation should be blocked in the constructor or not.

So if the users have the schema pairs available beforehand, they could enable the blocking codec generation when constructing fast reader/writers, and if the schema pairs are detected at runtime, the behavior should be disabled for more smooth user experience.

What do you think?

radai-rosenblatt:
so, for clarity, here's my proposed design for fast codecs (codec here means either encoder or decoder):

1. define a CodecPool abstraction. this would encapsulate the tread pool used to generate+compile codec classes, any custom classloader/caches, etc. this would allow controlling resources allocated to to code-gen - thread counts, priorities, memory, class caching, etc:
```java
interface CodecPool {
      //submits code-gen tasks immediately when called, task completes async via future
      CompletableFuture<Codec> generateCodec(Schema writer, Schema reader); 
}

AvroCompatibilityHelper {
    private final static CodecPool DEFAULT_POOL = ... //default. maybe lazily created on-demand
    
    CompletableFuture<Codec> buildFastCodec(@Nullable CodecPool pool, other args ...) {
        CodecPool pool = pool || DEFAULT_POOL;
        return pool.generateCodec(...other args)
    }
}
```
2. define a composite codec class to allow users to switch between vanilla avro and fast codecs if/when they are ready:
```java
class CompositeCodec implements Codec {
    private final Codec vanillaAvroCodec;
    private final CompletableFuture<Codec> fastCodecFuture;
    private final AtomicReference<Codec> codecToUse;

    public CompositeCodec(Codec vanillaAvroCodec, CompletableFuture<Codec> fastCodecFuture) {
       ...
       codecToUse = new ..(vanillaAvroCodec); //initially
       fastCodecFuture.addListener(() --> {codecToUse.set()}); //needs error handling
    }

    @Override
    public void doStuff() {
       codecToUse.get().doStuff()
    }
}
```
3. have a helper method to use the composite class:
```java
AvroCompatibilityHelper {
    public Codec buildLazyFastCodec(...) {
        return new CompositeCodec(normalAvro, buildFastCodec(...))
    }
}
```

users who want "warmup" can just do ```Helper.buildFastCodec().get(timeout)```. users who want to do I/O ASAP could use buildLazyFastCodec()

radai-rosenblatt (approved):
im ok with this under the understanding that as soon as time becomes available, this whole area of functionality gets refactored

yuzheng21:
close this PR. using a new clean one https://github.com/linkedin/avro-util/pull/249

PR #249, opened 2021-11-17
build and get fast deserializers async

yuzheng21:
- get fast deserializers async

Original PR https://github.com/linkedin/avro-util/pull/243 was polluted with rebased commits.

### DIFF
--- a/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDatumReaderWriterUtil.java
+++ b/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDatumReaderWriterUtil.java
@@ -2,6 +2,9 @@ package com.linkedin.avro.fastserde;
 
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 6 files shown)


radai-rosenblatt (approved):

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDatumReaderWriterUtil.java
```diff
@@ -55,6 +58,18 @@ public int hashCode() {
   private FastDatumReaderWriterUtil() {
   }
 
+  public static void warmUpFastGenericDatumReader(Schema writerSchema, Schema readerSchema, long timeoutInMillis)
```
radai-rosenblatt, line 61: this is ~trivial enough (getFitire().get(timeout) that we can expect users to write this directly in their code i think ?

yuzheng21: thanks. I removed these warm up utils.

radai-rosenblatt (approved):

PR #250, opened 2021-11-18
add code-gen options to fix/retain avro-702 impacted codegen

radai-rosenblatt:
add code transformation capability to replace SCHEMA$ contents to allow working around avro-702
also add codegen config option to specify which schemas NOT to fix, for backwards compatibility


srramach (approved):
LGTM. It would have been nice to add a test covering the "schemasToGenerateBadAvscFor" override.

PR #251, opened 2021-11-18
Optimize for very large schema

idanz:
When creating the code for a RECORD, in case the records contains more than 100 fields, the generator will start to create additional population functions, this is to prevent the generator from failing compiling the code with "code too large" error.


FelixGV:
Can you please add a test for this, and show what the code generated output is like? We have many tests that do it, so you can follow that pattern, on how to check-in the code gen changes for review.

In this case, since your schema and code gen output will likely be humongous, you might to make your threshold for when the special logic kicks in configurable. This way, you could, e.g., set the split threshold to 3 and code gen a serde for a schema with 8 fields and we should see two splits (IIUC).

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -280,16 +281,41 @@ private void processRecord(JVar recordSchemaVar, String recordName, final Schema
...
+
+    int fieldCount = 0;
+    int fieldsWrittenToTarget = 0;
+    JBlock popMethodBody = methodBody;
+    JMethod popMethod = null;
     for (Schema.Field field : recordWriterSchema.getFields()) {
+      // We roll the population method for very large records, the initial fields are kept in the outer method as original to maintain performance for smaller records
+      fieldCount++;
+      if (fieldCount % FIELDS_PER_POPULATION_METHOD == 0) {
+        popMethod = generatedClass.method(JMod.PUBLIC, codeModel.VOID,
```
gaojieliu, line 303: The method can be `private`, right?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -280,16 +281,41 @@ private void processRecord(JVar recordSchemaVar, String recordName, final Schema
...
+                getUniqueName("populate_" + recordName));
+
+        popMethod._throws(IOException.class);
+        if (recordAction.getShouldRead()) {
+          popMethod.param(recordClass, recordName);
```
gaojieliu, line 308: It is not clear to me whether the new smaller methods are using object reuse or not.
As @FelixGV mentioned, please commit the generated classes as well.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -310,12 +336,23 @@ private void processRecord(JVar recordSchemaVar, String recordName, final Schema
...
+        processSimpleType(field.schema(), readerFieldSchema, popMethodBody, action, putExpressionInRecord, fieldReuseSupplier);
+      }
+
+      if (popMethod != null) {
+        for(Class<? extends Exception> e: schemaAssistant.getExceptionsFromStringable()) {
+          popMethod._throws(e);
+        }
+      }
+
+      if (isTopLevel && (fieldsWrittenToTarget == recordReaderSchema.getFields().size())) {
```
gaojieliu, line 354: I am not sure whether this will be good for all the use cases since this behavior is different from Vanilla Avro.
If the input is a stream, which contains multiple records, and stop early will corrupt the deserialization of the following record.

gaojieliu (changes requested):

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
idanz: @gaojieliu can you give me example of such test that checks the actual generated code?
In terms of reuse, yes it should work fine since the reuse is use at the start to decide which object to populate and the smaller functions just do the actual population

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
idanz: how would you suggest to control this or any other parameter (like the split size) would it be by constructor or system property? currently I haven't seen much examples of configurables

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
idanz: yes sure

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
FelixGV: We should add tests for deserializing multiple sequential records from a single byte stream. This is a regular functionality of Avro and we do use it, so it's important that whatever efficiency improvement we build doesn't break anything. 

In this case, I don't think there should be any configs. Things must work out of the box no matter if we read just one record or multiple ones.

What you can do is to short-circuit your work as soon as possible, but still keep state which lets you know that if the user asks for one more record, then you still have this baggage from the end of the last record to parse and skip before starting to deserialize the next record.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
FelixGV: This is an example:

https://github.com/linkedin/avro-util/commit/c52235837db3954fd1c2a5671e0ccb4e5ed9f3ac

The commits before that are the changes related to this code gen.

idanz: But this is just the code gen, I haven't seen a test that actually validates the generated code itself, do you want me to just commit the generated code with a lower threshold? in that case I just wonder how I should configure the threshold? constructor?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
idanz: short circuit sounds very complicated, "marking" your position would be complicated and my bigger concern is that you need to remember the binary decoder but what do you do if you are given a different decoder in the next call? you might be in a multi-threaded situation. I think the caller must provide this on the deserialization command, any thoughts about that?

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
FelixGV: If you look at the commits preceding the one I linked, you will see the tests which validate that fast-avro works.

The goal of committing the code-gen is to be able to easily visualize the diff in code-gen output associated with a change in the code-gen logic. And we split off the code-gen output into a standalone commit to make it easier to sort through the logic changes versus code-gen output changes (the PR UI lets you pick a specific commit).

FelixGV: As for configurability, I leave that up to your judgment. A telescopic constructor, or system property or a package private setter which the test can access by being in the same package... any of these could do the trick.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
FelixGV: For sequential deserialization, there are some objects that keep getting reused -- at the very least the stream and I think the decoder too? There's got to be a way to keep a position somewhere...

idanz: yes but I think they are kept by the caller? I'm not familiar with the api, we call the "read" method directly with a binary decoder that we generate before hand

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
idanz: Hmm, I see a lot of tests for correctness of the deserializer but not something checking the code itself, maybe you can give me an example of such Test function? I have looked in the previous commits in the PR as well...
I will add the configuration anyway and I will make the tests work with a low value (e.g. 2) so all the functionality will be tested with this thing in place, however the default value for normal usage should probably be something like 100. 
In that case the generated code will indeed be different and I can commit that as well, would that be sufficient?

idanz:
Ok to make things more palatable I removed the 2nd optimization, so I'm just handling the "code too large" problem and it only kicks in when necessary so shouldn't cause issues for most.
I added a package-level modifier so the tests run with a low value of 2
I updated everything and regenerated the code for the tests so you can see the difference in an isolated commit.
I think that's everything you asked, please let me know if there's something else

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
FelixGV: The most important is that sequential deser of many records from the same stream keeps working, and we should add a test for that. Once we have a test, it'll be easy to validate if any given optimization can be adopted or not. If it breaks this use case, and we cannot find a way to make it work, then we should abandon it.

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
FelixGV: Yes, I think that's sufficient.

FelixGV:
> Ok to make things more palatable I removed the 2nd optimization, so I'm just handling the "code too large" problem and it only kicks in when necessary so shouldn't cause issues for most. I added a package-level modifier so the tests run with a low value of 2 I updated everything and regenerated the code for the tests so you can see the difference in an isolated commit. I think that's everything you asked, please let me know if there's something else

Thanks for your hard work, we'll take another look!

gaojieliu:
Hi @idanz 
Is it possible to re-arrange your commits into two commits instead of 6 in this PR as this one:
https://github.com/linkedin/avro-util/pull/233/commits

The first commit should only include code-gen logic change, and the second commit will only include all the generated files in `codegen` folder.
Also can you only apply the function breaking down logic in the new tests? you could specify a small limit and recover it after the test. With this way, the affected tests will be reduced a lot.

Right now, the PR is very difficult to review since it contains too many changes and spread across several different commits.

idanz:
@gaojieliu no problem to rearrange the commits however I don't think it's a good idea to limit the 2 to only "my" tests, when I ran everything with a size of 2 I found different features that didn't work well and had to update my code. And this might happen with future optimization. I think it's safer to apply this low limit on all tests for safety, what do you say?

FelixGV:
Sounds good, let's include the splitting behavior in various/all tests then. Consider that the splitting config can be an extra parameter permutation. We already have tests going through several permutations. Adding one more which doubles the quantity of tests is no big deal, since this test suite runs so fast anyway. Better be exhaustive.

idanz:
Ok I rebased the commits, are we good to go now?

gaojieliu (approved):

gaojieliu:
@idanz 
Cut a new release:
https://github.com/linkedin/avro-util/releases/tag/0.2.65

idanz:
Awesome, thanks!

idanz:
@gaojieliu I can't see the new version 0.2.65 in artifactory, but the snapshot was published for some reason

0.2.63/	19-11-21 08:02:58 +0700	
0.2.64/	21-11-21 09:48:23 +0700	
0.2.66-SNAPSHOT/

gaojieliu:
@radai-rosenblatt 
Do you have any idea about why ^?

gaojieliu:
Release log is here:
https://github.com/linkedin/avro-util/runs/4294091909?check_suite_focus=true

Problematic part:
```
> Task :publishAvro-fastserdePublicationToLinkedInJfrogRepository
Cannot upload checksum for module-maven-metadata.xml. Remote repository doesn't support sha-512. Error: Could not PUT 'https://linkedin.jfrog.io/artifactory/avro-util/com/linkedin/avroutil1/avro-fastserde/maven-metadata.xml.sha512'. Received status code 403 from server: Forbidden
```

radai-rosenblatt:
thats not the issue.

the issue is that not all modifications were committed or ignored, which leaves the working copy "dirty":
```
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization
	....more
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated
	...more
```

which causes this to actually release a snapshot version:
```
> Task :printVersion
0.2.67-SNAPSHOT
```

radai-rosenblatt:
@gaojieliu - either commit all the generated code or add it to gitignore. either way you want to arrive at a state where the build leaves a clean working copy

idanz:
@gaojieliu I created a new pull request to just add the missing generated files https://github.com/linkedin/avro-util/pull/256

PR #252, opened 2021-11-20
(badly) fix handling of escaped newlines and quotes in alternative avsc

radai-rosenblatt:
i've tried multiple StringEscapeUtils invocations to no avail, and as this is causing issues for me downstream im settling for simply nuking escaped newlines and quotes for now :-(


srramach (commented):
I recommend abandoning this PR and instead, let's try to fix #253. As mentioned in that issue, just nuking doc strings isn't enough (even as a band-aid). The same issue can happen with the `default` property as well.

radai-rosenblatt:
#254 properly fixes this

PR #254, opened 2021-11-20
[helper] Fix broken escaping in codegen by Avro 1.4/1.5.

### LINKED ISSUE #253: [helper] Handle improperly escaped chars in code generated by Avro 1.4/1.5.
Consider a valid schema that has embedded escaped chars in the `doc` property, like so:
```
{
  "type": "record",
  "name": "Foo",
  "fields": [{
    "name": "bar",
    "type": "string",
    "doc": "embedded \n escape \" chars"
  }]
}
```

If you try to codegen for this schema using Avro 1.4.* or Avro 1.5.*, the resulting code is broken and it looks like this:
```
... SCHEMA$ = ... "... \"embedded \n escape \\" chars\"}]}");
```

Notice how the `\\"` means that the `"` is no longer escaped, and terminates the Java string literal, making the rest of the line invalid Java.

This was fixed in Avro 1.6.0. See [this comment](https://issues.apache.org/jira/browse/AVRO-886?focusedCommentId=13105659&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13105659) by Doug Cutting on AVRO-886. The fix was effectively a single-line change in SpecificCompiler.java:
```
-    return o.toString().replace("\"", "\\\"");
+    return o.toString().replace("\\","\\\\").replace("\"", "\\\"");
```

I.e., instead of replacing only `"` with `\"`, the codegen also replaces all `\` with `\\` first.

We should do something equivalent in the compat helper's codegen.

I think it would be difficult to pre-process these schemas before handing them off to 1.4/1.5 for codegen. If we tried to do that, we'd replace only `\` with `\\` (and leave the replacement of `"` with `\"` to the native codegen). But this makes the schemas invalid JSON while they are in the intermediate state (where only `\` has been escaped). Which would mean native codegen wouldn't be able to parse the schemas to begin with.

Could we punt the problem by getting rid of the escaped chars completely from `doc`? After all, the doc string isn't used by readers or writers anyway. The problem is that this issue also occurs in the `default` property (which clearly cannot be arbitrarily modified without affecting readers). Note further that the default value can be a complex structure, with one or more strings anywhere within (e.g., a map, a deep record structure with strings somewhere within, etc).

/Cc: @radai-rosenblatt

srramach:
Avro 1.4/1.5 generate broken (uncompilable) code when the schema has
escaped chars in strings. We fix such code in the compat helper by:
* First, undoing the damage (unescaping `\"` back to `"`),
* Then, escaping backslashes (escaping `\` to `\\`), and
* Finally, re-escaping the quotes (from `"` to `\"`).

If we are replacing the schema with an "alternative", since the alt
comes directly from JSON, and not from codegen, we escape it using
a standard function (`StringEscapeUtils.escapeJava`), and don't bother
fixing up the original broken code.

If the code was generated by Avro 1.6 or later, it would be correctly
escaped to begin with, so there's nothing more to do.

The test was copied from PR #252, courtesy of @radai-rosenblatt.

Fixes #253.

### DIFF
--- a/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/CodeTransformations.java
+++ b/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/CodeTransformations.java
@@ -387,6 +387,13 @@ public class CodeTransformations {
     //drop in alternative avsc?
     if (alternativeAvsc != null && !alternativeAvsc.isEmpty()) {
       stringLiteral = StringEscapeUtils.escapeJava(alternativeAvsc);
+    } else if (generatedWith.earlierThan(AvroVersion.AVRO_1_6)) {
+      // See https://github.com/linkedin/avro-util/issues/253. Avro 1.4 and 1.5 generate broken code when there are
+      // escaped chars within strings. We undo the damage (unescape `\"` to `"`), and then fix it up properly (escape
+      // `\` to `\\`, followed by `"` to `\"`).
+      stringLiteral = stringLiteral.replace("\\\"", "\"")
+          .replace("\\", "\\\\")
+          .replace("\"", "\\\"");
     }
 
     boolean largeString = stringLiteral.length() >= MAX_STRING_LITERAL_SIZE;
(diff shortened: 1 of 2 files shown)


radai-rosenblatt (approved):

radai-rosenblatt:
great work!
thank you for investing the time effort and research into this!

PR #265, opened 2021-12-26
work on more customizable avsc generation and avro-702 mitigation

radai-rosenblatt:
this PR will add configuration options to the avsc generation method(s) of helper to allow for mitigation of avro-702.
planned changes:

- expend the configuration space for avsc generation to cover preferred/required use of vanilla avro, and "injecting" aliases to named types to better interop with their avro-702 mirror images 
- write code to inject avro-702 aliases in impacted schemas - add "correct" aliases when generating bad avsc, and "bad" aliases when generating correct avsc
- add tests demonstrating that alias injection above allows for good reader schemas to decode payloads written with bad schemas and vice versa (especially when enums and fixed types are impacted)

---
This change is [Reviewable](https://reviewable.io/reviews/linkedin/avro-util/265)

### DIFF
--- a/avro-fastserde/src/jmh/java/com/linkedin/avro/fastserde/generator/AvroRandomDataGenerator.java
+++ b/avro-fastserde/src/jmh/java/com/linkedin/avro/fastserde/generator/AvroRandomDataGenerator.java
@@ -16,7 +16,6 @@
 
 package com.linkedin.avro.fastserde.generator;
 
-import com.linkedin.avro.fastserde.generator.GenericRecordBuilder;
 import com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
@@ -28,7 +27,6 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Random;
 import org.apache.avro.Schema;
-import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericEnumSymbol;
 import org.apache.avro.generic.GenericFixed;
 import org.apache.avro.generic.GenericRecord;
@@ -38,6 +36,7 @@ import org.apache.avro.generic.GenericRecord;
  * Generates Java objects according to an {@link Schema Avro Schema}.
  */
 @SuppressWarnings("WeakerAccess")
+@Deprecated //functionality to be ported over to RandomRecordGenerator in "core" helper
 public class AvroRandomDataGenerator {
 
   /**
--- a/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvroAdapter.java
+++ b/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvroAdapter.java
@@ -129,7 +129,7 @@ public interface AvroAdapter {
     return Schema.createEnum(name, doc, namespace, values);
   }
 
-  String toAvsc(Schema schema, boolean pretty, boolean retainPreAvro702Logic);
+  String toAvsc(Schema schema, AvscGenerationConfig config);
 
   //code generation
 
--- a/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvscGenerationConfig.java
+++ b/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvscGenerationConfig.java
@@ -46,10 +46,22 @@ public class AvscGenerationConfig {
             true, false, true, Optional.of(Boolean.FALSE), false
     );
     /**
+     * always generates correct, pretty avsc, with aliases fo "bad" fullnames for better compatibility with avro 1.4
+     */
+    public static final AvscGenerationConfig CORRECT_MITIGATED_PRETTY = new AvscGenerationConfig(
+            false, false, true, Optional.of(Boolean.FALSE), true
+    );
+    /**
      * always generates correct, one-line avsc, with aliases fo "bad" fullnames for better compatibility with avro 1.4
      */
     public static final AvscGenerationConfig CORRECT_MITIGATED_ONELINE = new AvscGenerationConfig(
             false, false, false, Optional.of(Boolean.FALSE), true
+    );
+    /**
+     * always generates avro-702 impacted (but pretty) avsc. only use if you know what you're doing.
+     */
+    public static final AvscGenerationConfig LEGACY_PRETTY = new AvscGenerationConfig(
+            false, false, true, Optional.of(Boolean.TRUE), false
     );
     /**
      * always generates avro-702 impacted (but pretty) avsc, and adds aliases to the correct full names of
--- a/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvscGenerationConfig.java
+++ b/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvscGenerationConfig.java
@@ -1,0 +1,142 @@
+/*
+ * Copyright 2021 LinkedIn Corp.
+ * Licensed under the BSD 2-Clause License (the "License").
+ * See License in the project root for license information.
+ */
+
+package com.linkedin.avroutil1.compatibility;
+
+import org.apache.avro.Schema;
+
+import java.util.Optional;
+
+/**
+ * config class packing various configuration parameters and settings for the generation of
+ * avsc (json) strings out of {@link org.apache.avro.Schema} objects
+ */
+public class AvscGenerationConfig {
+    /**
+     * always punts to runtime avro to produce pretty output.
+     * result might suffer from @see <a href="https://issues.apache.org/jira/browse/AVRO-702">AVRO-702</a> under 1.4
+     * and also from incorrectly escaped json under versions smaller than 1.6
+     * (@see <a href="https://issues.apache.org/jira/browse/AVRO-886">AVRO-886</a>)
+     */
+    public static final AvscGenerationConfig VANILLA_PRETTY = new AvscGenerationConfig(
+            true, true, true, Optional.empty(), false
+    );
+    /**
+     * always punts to runtime avro to produce terse output.
+     * result might suffer from @see <a href="https://issues.apache.org/jira/browse/AVRO-702">AVRO-702</a> under 1.4
+     * and also from incorrectly escaped json under versions smaller than 1.6
+     * (@see <a href="https://issues.apache.org/jira/browse/AVRO-886">AVRO-886</a>)
+     */
+    public static final AvscGenerationConfig VANILLA_ONELINE  = new AvscGenerationConfig(
+            true, false, false, Optional.empty(), false
+    );
+    /**
+     * always generates correct, pretty-printed avsc. delegates this to runtime avro where possible
+     */
+    public static final AvscGenerationConfig CORRECT_PRETTY = new AvscGenerationConfig(
+            true, false, true, Optional.of(Boolean.FALSE), false
+    );
+    /**
+     * always generates correct, one-line avsc. delegates this to runtime avro where possible
+     */
+    public static final AvscGenerationConfig CORRECT_ONELINE = new AvscGenerationConfig(
+            true, false, true, Optional.of(Boolean.FALSE), false
+    );
+    /**
+     * always generates correct, one-line avsc, with aliases fo "bad" fullnames for better compatibility with avro 1.4
+     */
+    public static final AvscGenerationConfig CORRECT_MITIGATED_ONELINE = new AvscGenerationConfig(
+            false, false, false, Optional.of(Boolean.FALSE), true
+    );
+    /**
+     * always generates avro-702 impacted (but pretty) avsc, and adds aliases to the correct full names of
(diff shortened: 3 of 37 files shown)


srramach (changes requested):
> Reviewed 4 of 19 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/linkedin/avro-util/265)* status: 4 of 37 files reviewed, 9 unresolved discussions (waiting on @FelixGV and @radai-rosenblatt)

___
*[helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/Avro702Checker.java, line 14 at r2](https://reviewable.io/reviews/linkedin/avro-util/265#-Msa95Q_98CaMIepW8_T:-Msa95Q_98CaMIepW8_U:b4joovg) ([raw file](https://github.com/linkedin/avro-util/blob/7098d02f69ac61b71c86a35e1d32cd5741b8800d/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/Avro702Checker.java#L14)):*
> ```Java
>  * utility class that checks if schemas are susceptible to avro-702
>  */
> public class Avro702Checker {
> ```

Does this need to be public? Here and in other files below (such as AvroCodecUtil, etc), perhaps we can make the classes package scoped, so as not to expose these to users of the helper library? Otherwise, we run the risk that somebody is going to use them, forcing us to support these forever.



_Suggestion:_
```Java
class
```
___
*[helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/Avro702Checker.java, line 21 at r2](https://reviewable.io/reviews/linkedin/avro-util/265#-MsaNJd2-hJSwFmRFEey:-MsaNJd2-hJSwFmRFEez:b-q9e4sg) ([raw file](https://github.com/linkedin/avro-util/blob/7098d02f69ac61b71c86a35e1d32cd5741b8800d/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/Avro702Checker.java#L21)):*
> ```Java
> 
>     public static boolean isSusceptible(Schema schema) {
>         String badAvsc = AvroCompatibilityHelper.toAvsc(schema, AvscGenerationConfig.LEGACY_PRETTY);
> ```

Let's go with LEGACY_ONELINE here. It's not output or seen by any human, so why bother making it pretty?



_Suggestion:_
```Java
LEGACY_ONELINE
```
___
*[helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCodecUtil.java, line 31 at r2](https://reviewable.io/reviews/linkedin/avro-util/265#-Msa9g8H6sUA96zxzUAB:-Msa9g8H6sUA96zxzUAC:b-y9ezbq) ([raw file](https://github.com/linkedin/avro-util/blob/7098d02f69ac61b71c86a35e1d32cd5741b8800d/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCodecUtil.java#L31)):*
> ```Java
> public class AvroCodecUtil {
> 
>     public static byte[] serializeBinary(IndexedRecord record) throws Exception {
> ```

I think "throws Exception" is fine in unit tests, but not so in library code. Here and in other methods below, it's better to be more explicit about the types of exceptions that can be thrown.



_Suggestion:_
```Java
throws IOException, ...
```
___
*[helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java, line 985 at r2](https://reviewable.io/reviews/linkedin/avro-util/265#-MsaMwPF4ryTnKHdXsKI:-MsaMwPF4ryTnKHdXsKJ:bbz02ar) ([raw file](https://github.com/linkedin/avro-util/blob/7098d02f69ac61b71c86a35e1d32cd5741b8800d/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java#L985)):*
> ```Java
>             pretty ?
>                     AvscGenerationConfig.LEGACY_MITIGATED_PRETTY
>                   : AvscGenerationConfig.LEGACY_MITIGATED_ONELINE
> ```

Shouldn't we be using LEGACY_PRETTY and LEGACY_ONELINE here, to keep the previous semantics? If not, and you do want to return bad-but-mitigated schemas, you should fix the comment to say that the bad schema will still work.



_Code quote:_
```Java
                    AvscGenerationConfig.LEGACY_MITIGATED_PRETTY
                  : AvscGenerationConfig.LEGACY_MITIGATED_ONELINE
```
___
*[helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvscGenerationConfig.java, line 34 at r2](https://reviewable.io/reviews/linkedin/avro-util/265#-MsaCpsV5jBLGZrsmVAn:-MsaCpsV5jBLGZrsmVAo:bdn08oy) ([raw file](https://github.com/linkedin/avro-util/blob/7098d02f69ac61b71c86a35e1d32cd5741b8800d/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvscGenerationConfig.java#L34)):*
> ```Java
>      */
>     public static final AvscGenerationConfig VANILLA_ONELINE  = new AvscGenerationConfig(
>             true, false, false, Optional.empty(), false
> ```

The comment says "**always** punt to runtime avro", but the `force` argument is false. Shouldn't it be true?



_Suggestion:_
```Java
true
```
___
*[helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvscGenerationConfig.java, line 46 at r2](https://reviewable.io/reviews/linkedin/avro-util/265#-MsaDFz-19wukDcQVEPy:-MsaDFz-19wukDcQVEPz:b-fh3okk) ([raw file](https://github.com/linkedin/avro-util/blob/7098d02f69ac61b71c86a35e1d32cd5741b8800d/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvscGenerationConfig.java#L46)):*
> ```Java
>      */
>     public static final AvscGenerationConfig CORRECT_ONELINE = new AvscGenerationConfig(
>             true, false, true, Optional.of(Boolean.FALSE), false
> ```

The `pretty` argument should be false.



_Suggestion:_
```Java
false
```
___
*[helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvscGenerationConfig.java, line 65 at r2](https://reviewable.io/reviews/linkedin/avro-util/265#-MsaM9_E4u5HLbvRs-E1:-MsaM9_E4u5HLbvRs-E2:b-vg2lvl) ([raw file](https://github.com/linkedin/avro-util/blob/7098d02f69ac61b71c86a35e1d32cd5741b8800d/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvscGenerationConfig.java#L65)):*
> ```Java
>     public static final AvscGenerationConfig LEGACY_PRETTY = new AvscGenerationConfig(
>             false, false, true, Optional.of(Boolean.TRUE), false
>     );
> ```

Let's add a `LEGACY_ONELINE` for completeness.



_Code snippet:_
```Java
public static final AvscGenerationConfig LEGACY_ONELINE = new AvscGenerationConfig(
        false, false, false, Optional.of(Boolean.TRUE), false
);
```
___
*[helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvscGenerationConfig.java, line 74 at r2](https://reviewable.io/reviews/linkedin/avro-util/265#-MsaL6JoE9piE50Ng6SO:-MsaL6JoE9piE50Ng6SP:bya1blr) ([raw file](https://github.com/linkedin/avro-util/blob/7098d02f69ac61b71c86a35e1d32cd5741b8800d/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvscGenerationConfig.java#L74)):*
> ```Java
>     );
>     /**
>      * always generates avro-702 impacted (but pretty) avsc, and adds aliases to the correct full names of
> ```

Fix comment.



_Suggestion:_
```Java
impacted, one-line
```
___
*[helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvscGenerationConfig.java, line 92 at r2](https://reviewable.io/reviews/linkedin/avro-util/265#-MsaER-FEdmSrKFbjr4q:-MsaER-FEdmSrKFbjr4r:buqa99x) ([raw file](https://github.com/linkedin/avro-util/blob/7098d02f69ac61b71c86a35e1d32cd5741b8800d/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvscGenerationConfig.java#L92)):*
> ```Java
>      * not possible to delegate the work to vanilla runtime avro
>      */
>     private final boolean forceUseOfRuntimeAvro;
> ```

`(prefer: false, force: true)` is not a meaningful combo. It's good to avoid even the possibility of meaningless states, so how about folding both of these params into a single tri-state param, like `retainPreAvro702Logic`. I.e., `Optional<Boolean> useRuntimeAvro` with:
* false == always use our own logic (prefer: false, force: false)
* true == always use runtime Avro (prefer: true, force: true)
* empty == prefer runtime, but fallback to our logic where necessary (prefer: true, force: false)

In fact, thinking more about this, it feels like we shouldn't allow this to be selected by the caller at all. We should use either the runtime Avro or our own logic depending on what is asked for by the other params (pretty, pre702 and aliases). It would mean that for some combinations (e.g., Avro >= 1.7, aliases=false), we'll never use our own logic, but I think that's okay. No need to ever have our own implementation in those cases (and thus no need to test it, nor have a config that requests it).



_Suggestion:_
```Java
    private final Optional<Boolean> useRuntimeAvro;
```

radai-rosenblatt (commented):
> *[Reviewable](https://reviewable.io/reviews/linkedin/avro-util/265)* status: 4 of 37 files reviewed, 9 unresolved discussions (waiting on @FelixGV, @radai-rosenblatt, and @srramach)

___
*[helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/Avro702Checker.java, line 14 at r2](https://reviewable.io/reviews/linkedin/avro-util/265#-Msa95Q_98CaMIepW8_T:-MsbMgjm5TRh3vy_IIg5:bqyss2w) ([raw file](https://github.com/linkedin/avro-util/blob/7098d02f69ac61b71c86a35e1d32cd5741b8800d/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/Avro702Checker.java#L14)):*
<details><summary><i>Previously, srramach (Sreeram Ramachandran) wrote…</i></summary><blockquote>

Does this need to be public? Here and in other files below (such as AvroCodecUtil, etc), perhaps we can make the classes package scoped, so as not to expose these to users of the helper library? Otherwise, we run the risk that somebody is going to use them, forcing us to support these forever.
</blockquote></details>

I intend to use this as a utility in other projects, so yeah - public by design, not oversight
___
*[helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/Avro702Checker.java, line 21 at r2](https://reviewable.io/reviews/linkedin/avro-util/265#-MsaNJd2-hJSwFmRFEey:-MsbNF739hzqu-Dzxc5h:b-4jk7qu) ([raw file](https://github.com/linkedin/avro-util/blob/7098d02f69ac61b71c86a35e1d32cd5741b8800d/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/Avro702Checker.java#L21)):*
<details><summary><i>Previously, srramach (Sreeram Ramachandran) wrote…</i></summary><blockquote>

Let's go with LEGACY_ONELINE here. It's not output or seen by any human, so why bother making it pretty?
</blockquote></details>

it was being seen by me when i was debugging this, is why. fixed :-)
___
*[helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCodecUtil.java, line 31 at r2](https://reviewable.io/reviews/linkedin/avro-util/265#-Msa9g8H6sUA96zxzUAB:-MsbO5KwEm0vMlHd64dp:b-iob4xf) ([raw file](https://github.com/linkedin/avro-util/blob/7098d02f69ac61b71c86a35e1d32cd5741b8800d/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCodecUtil.java#L31)):*
<details><summary><i>Previously, srramach (Sreeram Ramachandran) wrote…</i></summary><blockquote>

I think "throws Exception" is fine in unit tests, but not so in library code. Here and in other methods below, it's better to be more explicit about the types of exceptions that can be thrown.
</blockquote></details>

IOException is the narrowest we can easily make this. in real life since we allocate and control buffers here even most of the IOExceptions are not really expected/possible, but some may come from malformed user payloads so im hesitant about going beyond that
___
*[helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java, line 985 at r2](https://reviewable.io/reviews/linkedin/avro-util/265#-MsaMwPF4ryTnKHdXsKI:-MsbOgWb0Y6ngJ8lVAcB:b-7ql1td) ([raw file](https://github.com/linkedin/avro-util/blob/7098d02f69ac61b71c86a35e1d32cd5741b8800d/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java#L985)):*
<details><summary><i>Previously, srramach (Sreeram Ramachandran) wrote…</i></summary><blockquote>

Shouldn't we be using LEGACY_PRETTY and LEGACY_ONELINE here, to keep the previous semantics? If not, and you do want to return bad-but-mitigated schemas, you should fix the comment to say that the bad schema will still work.
</blockquote></details>

fixed

PR #267, opened 2022-01-04
Change util method name from isGenericRecord to isGenericDataRecord and update its Javadoc accordingly

Lincong:
This util method is actually checking if a record is a `GenericData.Record` instead of an `org.apache.avro.generic.GenericRecord`. So, update its method name and doc to be more precise.


radai-rosenblatt:
could you please add back the original signature - mark it deprecated and have it call the new method (as-is this is breaking change)

radai-rosenblatt (approved):
thank you!

PR #269, opened 2022-01-13
[helper] Set the field default sort order to ASCENDING.

### LINKED ISSUE #268: [helper] FieldBuilder causes NPE when order is not supplied for a field
I have some code that looks like:
```
AvroCompatibilityHelper.newField(null)
  .setName("foo")
  .setSchema(...)
  .build()
```

This throws the following exception:
```
java.lang.NullPointerException
	at java.util.Objects.requireNonNull(Objects.java:203)
	at org.apache.avro.Schema$Field.<init>(Schema.java:597)
	at com.linkedin.avroutil1.compatibility.avro110.FieldBuilder110.build(FieldBuilder110.java:101)
```

The line in question from `Schema$Field` is the `Objects.requireNonNull` line below:
```java
    public Field(String name, Schema schema, String doc, Object defaultValue, Order order) {
      this(name, schema, doc,
          defaultValue == NULL_DEFAULT_VALUE ? NullNode.getInstance() : JacksonUtils.toJsonNode(defaultValue), true,
          Objects.requireNonNull(order));
    }
```
The issue appears to come from this line in `FieldBuilder110`:
```java
    Schema.Field result = new Schema.Field(_name, _schema, _doc, avroFriendlyDefault, _order);
```
Overloads exist for `new Schema.Field` which do not require the order to be specified, and it defaults to `Order.ASCENDING`. I would expect the builder to behave similarly, by not requiring order and just defaulting to `Order.ASCENDING` if the user did not supply it.

This is on Avro 1.10, I am not sure if it applies to other Avro versions as well.

srramach:
Fixes #268.

### DIFF
--- a/helper/impls/helper-impl-110/src/main/java/com/linkedin/avroutil1/compatibility/avro110/FieldBuilder110.java
+++ b/helper/impls/helper-impl-110/src/main/java/com/linkedin/avroutil1/compatibility/avro110/FieldBuilder110.java
@@ -98,7 +98,7 @@ public class FieldBuilder110 implements FieldBuilder {
     } catch (Exception e) {
       throw new IllegalArgumentException("unable to convert default value " + _defaultVal + " into something avro can handle", e);
     }
-    Schema.Field result = new Schema.Field(_name, _schema, _doc, avroFriendlyDefault, _order);
+    Schema.Field result = new Schema.Field(_name, _schema, _doc, avroFriendlyDefault, _order == null ? Order.ASCENDING : _order);
     if (_props != null) {
       for (Map.Entry<String, Object> entry : _props.entrySet()) {
         result.addProp(entry.getKey(), entry.getValue());
--- a/helper/impls/helper-impl-110/src/main/java/com/linkedin/avroutil1/compatibility/avro110/FieldBuilder110.java
+++ b/helper/impls/helper-impl-110/src/main/java/com/linkedin/avroutil1/compatibility/avro110/FieldBuilder110.java
@@ -19,7 +19,7 @@ public class FieldBuilder110 implements FieldBuilder {
   private Schema _schema;
   private String _doc;
   private Object _defaultVal;
-  private Order _order;
+  private Order _order = Order.ASCENDING;
   private Map<String, Object> _props;
 
   public FieldBuilder110(Schema.Field other) {
@@ -28,7 +28,7 @@ public class FieldBuilder110 implements FieldBuilder {
       _schema = other.schema();
       _doc = other.doc();
       _defaultVal = other.defaultVal();
-      _order = other.order();
+      _order = other.order();  // other.order() cannot be null under Avro 1.10.*
       _props = other.getObjectProps();
     }
   }
@@ -80,6 +80,9 @@ public class FieldBuilder110 implements FieldBuilder {
 
   @Override
   public FieldBuilder setOrder(Order order) {
+    if (order == null) {
+      throw new IllegalArgumentException("sort order cannot be null");
+    }
     _order = order;
     return this;
   }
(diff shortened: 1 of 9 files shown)


helper/impls/helper-impl-110/src/main/java/com/linkedin/avroutil1/compatibility/avro110/FieldBuilder110.java
```diff
@@ -98,7 +98,7 @@ public FieldBuilder copyFromField() {
     } catch (Exception e) {
       throw new IllegalArgumentException("unable to convert default value " + _defaultVal + " into something avro can handle", e);
     }
-    Schema.Field result = new Schema.Field(_name, _schema, _doc, avroFriendlyDefault, _order);
+    Schema.Field result = new Schema.Field(_name, _schema, _doc, avroFriendlyDefault, _order == null ? Order.ASCENDING : _order);
```
radai-rosenblatt, line 101: nit - if we initialize the _order field to ASCENDING in the parent class it would save us having to do this in all the subclasses ?

radai-rosenblatt: (we would need to add an assert that _order != null at this point though)

radai-rosenblatt: ...which maybe we can also do in the parent class by introducing some validate() method called 1st thing on build()

srramach: I wanted to do it this way so that even if somebody explicitly calls `setOrder(null)`, we'll correctly transform it into ASCENDING.

helper/tests/helper-tests-allavro/src/test/java/com/linkedin/avroutil1/compatibility/FieldBuilderTest.java
```diff
@@ -67,6 +67,39 @@ public void testDefaultValuesInterop() throws Exception {
...
+        }
+    }
+
+    @Test
+    public void testNewFieldSortOrder() throws Exception {
+        // default (no order specified).
+        Schema.Field field = AvroCompatibilityHelper.newField(null).setName("default").build();
+        Assert.assertEquals(field.order(), Schema.Field.Order.ASCENDING);
+
+        // explicit set the order to null (in our API); should get transformed to ASCENDING.
```
radai-rosenblatt, line 86: should result in IllegalArgumentException out of build(). likely users who set to null are doing something wrong?

helper/impls/helper-impl-110/src/main/java/com/linkedin/avroutil1/compatibility/avro110/FieldBuilder110.java
srramach: Also, there's no "parent class". Each of the FieldBuilderXX implements an interface, and don't extend some parent class.

srramach: Changed the code. PTAL.

helper/tests/helper-tests-allavro/src/test/java/com/linkedin/avroutil1/compatibility/FieldBuilderTest.java
srramach: Done, but doing it in setOrder(), as discussed offline.

radai-rosenblatt (approved):

PR #272, opened 2022-01-24
rename cloneSchema to newSchema to align with API for field builders, fix bug in CodecUtil

radai-rosenblatt:


helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AbstractSchemaBuilder.java
```diff
@@ -23,13 +23,24 @@
...
+        if (original != null) {
+            _type = original.getType();
+            _name = original.getName();
+            _namespace = original.getNamespace();
+            _doc = original.getDoc();
+            _isError = original.isError();
+            //make a copy of fields so its mutable
+            _fields = new ArrayList<>(original.getFields());
+        } else {
+            _fields = new ArrayList<>(1);
```
srramach, line 35: Why 1 and not 0?

srramach (approved):

helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AbstractSchemaBuilder.java
radai-rosenblatt: micro optimization. i assume the schema to be created will be a record, and will have at least a single field. given that the schema doesnt even have to be a record maybe i'll start fields out as null ?

srramach: Yeah, I guess either null or empty list is more intuitive. I am fine with 1 as well, but in that case, please add a comment explaining what you just said above; it's a non-intuitive default.

PR #282, opened 2022-02-16
[helper] Transform null values within default values into null marker objects.

srramach:
When transforming a default value that has a map somewhere inside, and
when that map has null values in it, we need to transform those nulls
into actual objects. Otherwise, Avro croaks like so:
```
java.lang.NullPointerException
    at org.apache.avro.util.internal.JacksonUtils.toJson(JacksonUtils.java:96)
    at org.apache.avro.util.internal.JacksonUtils.toJson(JacksonUtils.java:68)
    at org.apache.avro.util.internal.JacksonUtils.toJsonNode(JacksonUtils.java:53)
    at org.apache.avro.Schema$Field.<init>(Schema.java:596)
    at com.linkedin.avroutil1.compatibility.avro110.FieldBuilder110.build(FieldBuilder110.java:104)
    at com.linkedin.avroutil1.compatibility.FieldBuilderTest.setDefaultValues(FieldBuilderTest.java:118)
    at com.linkedin.avroutil1.compatibility.FieldBuilderTest.testDefaultValuesInterop(FieldBuilderTest.java:68)
```

Schema.Field's constructor calls JacksonUtils.toJsonNode(defaultValue).
If the defaultValue is itself null, it's okay. From JacksonUtils.java:
```
public static JsonNode toJsonNode(Object datum) {
  if (datum == null) {
    return null;
  }
  ...
```

But if the defaultValue is not null, it descends into the toJson()
method, which doesn't expect to see null anywhere within. When it sees
anything it doesn't recognize, it tries to throw AvroRuntimeException,
but the call to datum.getClass() (to construct the message for the
exception) causes the NPE.

The correct way to resolve this involves transforming nulls into
JsonProperties.NULL_VALUE. That only exists in Avro >= 1.8, so the
avroFriendlyDefaultValue() method has to be hoisted out of common and
into each of the relevant FieldBuilderNN classes, resulting in some
code duplication.

### DIFF
--- a/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvroSchemaUtil.java
+++ b/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvroSchemaUtil.java
@@ -8,15 +8,10 @@ package com.linkedin.avroutil1.compatibility;
 
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
-import java.util.HashMap;
 import java.util.IdentityHashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.stream.Collectors;
 import org.apache.avro.Schema;
-import org.apache.avro.generic.GenericData;
-import org.apache.avro.generic.GenericFixed;
-import org.apache.avro.generic.IndexedRecord;
 
 
 public class AvroSchemaUtil {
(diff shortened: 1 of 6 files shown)


srramach:
I forgot to add that this PR removes a public API but I don't believe anybody else is using it yet, so shouldn't break anything.

radai-rosenblatt (approved):

PR #286, opened 2022-02-22
flesh out SchemaBuilders to cover more than just record schemas

radai-rosenblatt:


radai-rosenblatt:
differences across versions so far deal with props, defaults (which are not yet supported) and enum defaults.

reason i dont have more specific builder classes is the amount of effort vs the limited gain. also if you look at avro itself and its "fancy" builder syntax i think the end result is actually worse - its very confusing (for me?) to use

srramach:
> differences across versions so far deal with props, defaults (which are not yet supported) and enum defaults.

Okay. I _think_ it's possible to implement them using only Avro 1.4 APIs (or API hacks, such as setAccessible(true)), such that a single implementation would work under any runtime Avro version. But I haven't explored it fully. I'll try to do that in my spare time, and if it does work, we can get rid of all the SchemaBuilderNN impls. tl;dr: Don't worry about it for now; let's see if I can make it happen.

> reason i dont have more specific builder classes is the amount of effort vs the limited gain. also if you look at avro itself and its "fancy" builder syntax i think the end result is actually worse - its very confusing (for me?) to use

I agree that the Avro APIs for schema builders is very wonky. But, I think we can have the best of both worlds? So let's say you have this in your current model:
```
// Represents:
// { "type": "record":
//   "fields": [
//       { "type": "fixed", "size": 4 },
//       { "type": "record",
//         "fields": [
//              { "type": "string", "default": "foo" }
// ...
new SchemaBuilderNN().addField(FIXED).setSize(4).addField(RECORD).addField(STRING).setDefault("foo")...
```
In my proposed model, the main concrete builder would still have the same API as above, but would defer to individual subclasses as needed:
```
new RecordBuilder().addField(FIXED).setSize(4).addField(RECORD).addField(STRING).setDefault("foo")...

public abstract class AbstractSchemaBuilder {
    private AbstractSchemaBuilder parent;
    public abstract AbstractSchemaBuilder addField(Type type) { throw new UnsupportedOperationException(...); }
    public abstract AbstractSchemaBuilder setSize(int size) { throw new UnsupportedOperationException(...); }
    ...
}

class RecordSchemaBuilder extends AbstractSchemaBuilder {
    private List<AbstractSchemaBuilder> fields_;
    @Override
    public AbstractSchemaBuilder addField(Type type) {
        AbstractSchemaBuilder child;
        switch (type) {
            case RECORD: child = new RecordSchemaBuilder(...); break;
            case FIXED: child = new FixedSchemaBuilder(...); break;
            ...
        }
        child.parent = this;
        return child;
    }
    ...  // Does not implement setSize().
}

class FixedSchemaBuilder extends AbstractSchemaBuilder {
    private int size_;
    @Override
    public AbstractSchemaBuilder addField(Type type) {
        return parent.addField(type);  // We are not a record, so delegate to the parent.
    }

    @Override
    public AbstractSchemaBuilder setSize(int size) {
        size_ = size;
    }
    ...
}
```

Could that work?

radai-rosenblatt:
i've simplified the PR a little (not much).
i dont know how feasible the fancier proposed API is, given the need to support:

- schema cloning (start builder providing another schema) - the return value will have to be the parent API forcing users to cast?
- default values - another area where avro support varied wildly across versions
- logical types (if we ever get there) - would force you to define a builder interface per avro type almost?

PR #291, opened 2022-03-03
Added getCompatibleDefaultValue method and un-deprecation of createSchemaField

abhisheknath2011:
1. Added getNullableGenericDefaultValue to check if schema field has default value. If so returns the default value otherwise returns null. So, null means that the field does not have default value or field is of type NULL or union field with null and null being the first member.
2. Un deprecated createSchemaField as it is easier to invoke compared to newField if there there are more usage within a file or within a single method. This will help make code more readable and reduce the code size.
3. Added some test for the newly added methods.

### DIFF
--- a/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
+++ b/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
@@ -603,9 +603,9 @@ public class AvroCompatibilityHelper {
    * @param field a schema field
    * @return the default value of the field (if such a value exists),
    *         as a generic record class. Returns null if there is no default value.
-   *         null default also means union field with null. If null default value
-   *         is returned, if needed choose appropriate default value based on
-   *         schema type.
+   *         A null default also means that field is of type NULL or union field with null 
+   *         and null being the first member. If null default value is returned, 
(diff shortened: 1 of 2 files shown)


helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
```diff
@@ -595,6 +595,22 @@ public static Object getGenericDefaultValue(Schema.Field field) {
...
 
+  /**
+   * returns the default value for a schema field, as a generic record class
+   * (if the default value is complex enough - say records, enums, fixed fields etc)
+   * or as a JDK/Avro class (for simple values like Strings or booleans). <br>
+   *
+   * @param field a schema field
+   * @return the default value of the field (if such a value exists),
+   *         as a generic record class. Returns null if there is no default value.
+   *         null default also means union field with null. If null default value
```
radai-rosenblatt, line 606: null default could technically mean the field is of type NULL (and not a union).
i think its enough to explain that a return value of null could either mean field has no default or that the field default is null

abhisheknath2011: I have updated the description based on your comment. I think it would be more clear to include the union case as well.

radai-rosenblatt (approved):

PR #301, opened 2022-03-22
Added spotbugs rule to detect usage of org.apache.avro.Schema.toString usage

li-ukumar:
org.apache.avro.Schema.toString is vulnerable to AVRO-702 in avro v1.4
Added spotbugs to detect and notify

### DIFF
--- a/spotbugs-plugin/src/main/java/com/linkedin/avroutil1/spotbugs/SchemaSerializationDetector.java
+++ b/spotbugs-plugin/src/main/java/com/linkedin/avroutil1/spotbugs/SchemaSerializationDetector.java
@@ -9,6 +9,7 @@ package com.linkedin.avroutil1.spotbugs;
 import edu.umd.cs.findbugs.BugInstance;
 import edu.umd.cs.findbugs.BugReporter;
 import edu.umd.cs.findbugs.bcel.OpcodeStackDetector;
+import javax.annotation.Nonnull;
 import org.apache.bcel.Const;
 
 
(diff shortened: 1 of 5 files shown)


spotbugs-plugin/src/main/java/com/linkedin/avroutil1/spotbugs/SchemaSerializationDetector.java
```diff
@@ -0,0 +1,50 @@
...
+  public SchemaSerializationDetector(BugReporter bugReporter) {
+    this.bugReporter = bugReporter;
+  }
+
+  @Override
+  public void sawOpcode(int seen) {
+    if (seen != Const.INVOKEVIRTUAL && seen != Const.INVOKESTATIC) {
+      return;
+    }
+    if (getClassConstantOperand().equals("org/apache/avro/Schema") && getMethodDescriptorOperand().getName()
```
radai-rosenblatt, line 35: getClassConstantOperand() is evaluated before getMethodDescriptorOperand().getName().equals("toString").

and it could be null.

you should have an outer condition on the method name and only then check 
 getClassConstantOperand(), and also handle the possiblity of getClassConstantOperand() returning null (in iwhich case continue)

radai-rosenblatt (approved):

spotbugs-plugin/src/main/java/com/linkedin/avroutil1/spotbugs/SchemaSerializationDetector.java
li-ukumar: Done

PR #308, opened 2022-04-06
[AvscWriter] enforce uniform numeric defaults on write

### LINKED ISSUE #289: fix avscwriter's handling of numeric default values
avro 1.9 in particular suffers from issues around numeric default values - https://issues.apache.org/jira/browse/AVRO-2648

we need the output of avscwriter to be well defined and the same under all runtime versions of avro.

the correct thing to do is probably to force default values to be floating point json literals for floating point types (so "0.0") and integer json literals for integer types.

dg-builder:
fixes #289

### Expected functionality for default values:
For `int`s and `long`s, integer values that are represented as `float`s/`double`s (i.e. `-1.0`, `0.0`) will be converted to an `int`/`long` (i.e. `-1`, `0`). If the number is a true double (i.e. `3.14159`) we will warn and passthrough.

For `float`s and `double`s, integer values (i.e. `-1`, `0`) are represented with a decimal point and a trailing 0 (i.e. `-1.0`, `0.0`).

### DIFF
--- a/helper/impls/helper-impl-110/src/main/java/com/linkedin/avroutil1/compatibility/avro110/Avro110AvscWriter.java
+++ b/helper/impls/helper-impl-110/src/main/java/com/linkedin/avroutil1/compatibility/avro110/Avro110AvscWriter.java
@@ -27,6 +27,8 @@ import java.util.Set;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.apache.avro.Schema.Type.*;
+
 
 public class Avro110AvscWriter extends AvscWriter<Jackson2JsonGeneratorWrapper> {
     private static final JsonFactory FACTORY = new JsonFactory().setCodec(new ObjectMapper());
@@ -106,7 +108,11 @@ public class Avro110AvscWriter extends AvscWriter<Jackson2JsonGeneratorWrapper> 
     private JsonNode enforceUniformNumericDefaultValues(Schema.Field field) {
         JsonNode genericDefaultValue = Accessor.defaultValue(field);
         double numericDefaultValue = genericDefaultValue.doubleValue();
-        switch (field.schema().getType()) {
+        Schema schema = field.schema();
+        // a default value for a union, must match the first element of the union
+        Schema.Type defaultType = schema.getType() == UNION ? schema.getTypes().get(0).getType() : schema.getType();
+
+        switch (defaultType) {
             case INT:
                 if (numericDefaultValue % 1 != 0) {
                     LOGGER.warn(String.format("Invalid default value: %s for \"int\" field: %s", genericDefaultValue, field.name()));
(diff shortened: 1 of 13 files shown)


helper/impls/helper-impl-110/src/main/java/com/linkedin/avroutil1/compatibility/avro110/Avro110AvscWriter.java
```diff
@@ -85,4 +96,36 @@ private void writeProps(Map<String, Object> props, Jackson2JsonGeneratorWrapper
             delegate.writeObjectField(entry.getKey(), JacksonUtils.toJsonNode(o));
         }
     }
+
+    /**
+     *  Enforces uniform numeric default values across Avro versions
+     */
+    private JsonNode enforceUniformNumericDefaultValues(Schema.Field field) {
+        JsonNode genericDefaultValue = Accessor.defaultValue(field);
+        if (!genericDefaultValue.isNumber()) {
```
radai-rosenblatt, line 105: will this complain and "abort" for fields that are not numeric types?

helper/impls/helper-impl-110/src/main/java/com/linkedin/avroutil1/compatibility/avro110/Avro110AvscWriter.java
```diff
@@ -85,4 +96,36 @@ private void writeProps(Map<String, Object> props, Jackson2JsonGeneratorWrapper
...
+            LOGGER.warn(String.format("Invalid default value: %s for \"long\" field: %s", genericDefaultValue, field.name()));
+            return genericDefaultValue;
+        }
+        double numericDefaultValue = genericDefaultValue.doubleValue();
+        switch (field.schema().getType()) {
```
radai-rosenblatt, line 110: fields can be unions?

helper/impls/helper-impl-110/src/main/java/com/linkedin/avroutil1/compatibility/avro110/Avro110AvscWriter.java
dg-builder: Ah that's right. I've moved the `isNumber` so that it preceeds the call to `enforceUniformNumericDefaultValues`

helper/impls/helper-impl-110/src/main/java/com/linkedin/avroutil1/compatibility/avro110/Avro110AvscWriter.java
dg-builder: should be fixed in the most recent commit

settings.gradle
```diff
@@ -23,7 +23,7 @@ rootProject.name = 'avro-util'
 include 'test-common'
 include 'parser'
 include 'avro-codegen'
-include 'avro-fastserde'
+//include 'avro-fastserde'
```
radai-rosenblatt, line 26: oops

radai-rosenblatt (approved):

Issue #271, opened 2022-01-18
detect version mismatches between avro and avro-compiler


radai-rosenblatt:
here's avro-compiler 1.9 with "core" avro 1.10 failing to find a class that's been deleted:
```
Exception in thread "main" java.lang.IllegalStateException: java.lang.reflect.InvocationTargetException
        at com.linkedin.avroutil1.compatibility.avro110.Avro110Adapter.compile(Avro110Adapter.java:343)
        at com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper.compile(AvroCompatibilityHelper.java:561)
        at com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper.compile(AvroCompatibilityHelper.java:569)
        at org.apache.avro.specific.SpecificCompilerHelper.compile(SpecificCompilerHelper.java:37)
        at com.linkedin.avro.Builder.main(Builder.java:209)
Caused by: java.lang.reflect.InvocationTargetException
        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
        at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
        at com.linkedin.avroutil1.compatibility.avro110.Avro110Adapter.compile(Avro110Adapter.java:307)
        ... 4 more
Caused by: java.lang.NoClassDefFoundError: org/apache/avro/data/JodaTimeConversions$DateConversion
```

PR #316, opened 2022-04-20
adds detectAvroCompilerVersion() and accompanying tests

dg-builder:
fixes https://github.com/linkedin/avro-util/issues/271

### DIFF
--- a/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
+++ b/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
@@ -836,7 +836,7 @@ public class AvroCompatibilityHelper {
   private static AvroVersion detectAvroCompilerVersion() {
     Class<?> specificCompilerClass;
 
-    // In Avro 1.5.0+, SpecificRecord is part of the org.apache.avro.compiler.specific ClassPath.
+    // In Avro 1.5.0+, SpecificCompiler is part of the org.apache.avro.compiler.specific package.
     try {
       specificCompilerClass = Class.forName("org.apache.avro.compiler.specific.SpecificCompiler");
     } catch (ClassNotFoundException expected) {
@@ -846,14 +846,11 @@ public class AvroCompatibilityHelper {
         Class.forName("org.apache.avro.specific.SpecificCompiler");
         return AvroVersion.AVRO_1_4;
 
-        // In the case that SpecificRecord doesn't exist in the ClassPath at all (only possible in 1.5+).
+        // In the case that SpecificCompiler doesn't exist in the ClassPath at all (only possible in 1.5+).
       } catch (ClassNotFoundException ignored) {
         return null;
       }
     }
-
-
-
 
     // SpecificCompiler.isUnboxedJavaTypeNullable(Schema s) method was added to 1.6.0.
     try {
@@ -869,14 +866,10 @@ public class AvroCompatibilityHelper {
     // of the field only indicates that version is one of 1.6, 1.10+.
 
     // SpecificCompiler.hasBuilder(Schema s) method was added in 1.8.0
-    Boolean fileResourceLoaderClassExists = false;
     try {
-      fileResourceLoaderClassExists = checkIfVelocityEngineFileResourceLoaderClassExists(specificCompilerClass);
       specificCompilerClass.getMethod("hasBuilder", Schema.class);
-    } catch (IllegalStateException unexpected) {
-      return null;
     } catch (NoSuchMethodException expected) {
-      if (fileResourceLoaderClassExists) {
+      if (checkIfVelocityEngineFileResourceLoaderClassExists(specificCompilerClass)) {
         return AvroVersion.AVRO_1_7;
       } else {
         return AvroVersion.AVRO_1_6;
@@ -913,7 +906,7 @@ public class AvroCompatibilityHelper {
       throws IllegalStateException {
     try {
       // If `velocityEngine` were public, we could simplify this reflection to be
-      // ```VelocityEngine velocityEngine = (new SpecificRecord()).velocityEngine;````
+      // ```VelocityEngine velocityEngine = (new SpecificCompiler()).velocityEngine;````
       Field velocityEngineField = specificCompilerClass.getDeclaredField("velocityEngine");
       velocityEngineField.setAccessible(true);
       // SpecificCompiler's default constructor is also private.
@@ -936,7 +929,6 @@ public class AvroCompatibilityHelper {
           expected.fillInStackTrace());
     }
   }
-
 
   /**
    * Get the full schema name for records or type name for primitives. This adds compatibility
--- a/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
+++ b/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
@@ -23,7 +23,6 @@ import java.io.ObjectOutput;
 import java.io.OutputStream;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
-import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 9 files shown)


helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
```diff
@@ -54,9 +62,16 @@ public class AvroCompatibilityHelper {
...
     DETECTED_VERSION = detectAvroVersion();
+    DETECTED_COMPILER_VERSION = detectAvroCompilerVersion();
     if (DETECTED_VERSION == null) {
       ADAPTER = null;
     } else {
+      if (DETECTED_VERSION != DETECTED_COMPILER_VERSION) {
+        LOGGER.error(
+            String.format("avro and avro-compiler version mismatch! avro version: %s. avro-compiler version: %s",
+                DETECTED_VERSION, DETECTED_COMPILER_VERSION));
+      }
```
dg-builder, line 73: Not sure if we want to log or throw here...

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
```diff
@@ -100,6 +115,14 @@ public static AvroVersion getRuntimeAvroVersion() {
...
   }
 
+  /**
+   * returns the detected runtime version of avro-compiler, or null if none found
+   * @return the version of avro detected on the runtime classpath, or null if no avro found
+   */
+  public static AvroVersion getRuntimeAvroCompilerVersion() {
+    return DETECTED_COMPILER_VERSION;
+  }
+
```
dg-builder, line 125: This is only public in order to test against this method. Not sure if there is a better way without using Guava's `@VisibleForTesting`

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
```diff
@@ -814,6 +837,93 @@ private static AvroVersion detectAvroVersion() {
...
+      // Without reflection, this would be ```velocityEngine.getProperty("file.resource.loader.class");```
+      Class<?> velocityEngineClass = Class.forName("org.apache.velocity.app.VelocityEngine");
+      Method velocityEngineGetPropertyMethod = velocityEngineClass.getMethod("getProperty", String.class);
+      Object fileResourceLoaderClassProperty = velocityEngineGetPropertyMethod.invoke(velocityEngine, "file.resource.loader.class");
+
+      return fileResourceLoaderClassProperty != null;
+
+      // We should never reach these exceptions
+    } catch (InstantiationException | IllegalAccessException | NoSuchMethodException | ClassNotFoundException | InvocationTargetException | NoSuchFieldException expected) {
+      throw new RuntimeException(expected.fillInStackTrace());
```
dg-builder, line 922: lmk if there's a better exception to raise here...

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
radai-rosenblatt: throw, definitely throw.
the most lenient which i still think is reasonable is only throw if/when any compiler-related methods are called, but to begin with i'd just throw.

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
```diff
@@ -100,6 +115,14 @@ public static AvroVersion getRuntimeAvroVersion() {
     return DETECTED_VERSION;
   }
 
+  /**
+   * returns the detected runtime version of avro-compiler, or null if none found
+   * @return the version of avro detected on the runtime classpath, or null if no avro found
```
radai-rosenblatt, line 120: nit "if no avro-compiler found"

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
```diff
@@ -814,6 +837,95 @@ private static AvroVersion detectAvroVersion() {
     return AvroVersion.AVRO_1_11;
   }
 
+  private static AvroVersion detectAvroCompilerVersion() {
+    Class<?> specificCompilerClass;
+
+    // There is no avro-compiler in 1.4.X
+    try {
+      specificCompilerClass = Class.forName("org.apache.avro.compiler.specific.SpecificCompiler");
+    } catch (ClassNotFoundException unexpected) {
```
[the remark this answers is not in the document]
radai-rosenblatt: org.apache.avro.specific.SpecificCompiler is the FQCN of the 1.4 class, where all newer versions have org.apache.avro.compiler.specific package - so wont it be better to also positively check for org.apache.avro.specific.SpecificCompiler ?

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
```diff
@@ -814,6 +837,95 @@ private static AvroVersion detectAvroVersion() {
...
+    } catch (NoSuchMethodException expected) {
+      return AvroVersion.AVRO_1_5;
+    }
+
+    // In AVRO-1094, a new property was added to velocityEngine in version 1.7.0.
+    // There were no differences in the public interface, so we check for the presence of a property in the private
+    // field, `velocityEngine`.
+    // To further complicate matters, this field is used in 1.7.0 and removed by 1.10.2. So the presence
+    // of the field only indicates that version is one of 1.6, 1.10+.
+    boolean fileResourceLoaderClassExists = checkIfVelocityEngineFileResourceLoaderClassExists(specificCompilerClass);
```
radai-rosenblatt, line 862: this computation should move into the catch block, because in some future >-- 1.11 avro version it might throw, and you only need this value if the version is < 1.8

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
radai-rosenblatt: i recommend throwing something like IllegalStateException with the msg "unexpected exception while retrieving velocity engine config" and supplying the caught exception as a cause.

radai-rosenblatt: you should also be able to simplify this to catch(Exception) {} i think ?

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
dg-builder: Good point, I'll leave it as `public`

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
```diff
@@ -814,6 +833,111 @@ private static AvroVersion detectAvroVersion() {
     return AvroVersion.AVRO_1_11;
   }
 
+  private static AvroVersion detectAvroCompilerVersion() {
+    Class<?> specificCompilerClass;
+
+    // In Avro 1.5.0+, SpecificRecord is part of the org.apache.avro.compiler.specific ClassPath.
```
radai-rosenblatt, line 839: nit - SpecificCOmpiler?

radai-rosenblatt: also classpath --> package

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
```diff
@@ -814,6 +833,111 @@ private static AvroVersion detectAvroVersion() {
...
+    try {
+      specificCompilerClass = Class.forName("org.apache.avro.compiler.specific.SpecificCompiler");
+    } catch (ClassNotFoundException expected) {
+      try {
+        // In Avro 1.4, SpecificCompiler was a class of org.apache.avro.specific. In 1.5.0, SpecificCompiler was moved
+        // to its own package.
+        Class.forName("org.apache.avro.specific.SpecificCompiler");
+        return AvroVersion.AVRO_1_4;
+
+        // In the case that SpecificRecord doesn't exist in the ClassPath at all (only possible in 1.5+).
```
radai-rosenblatt, line 849: nit - SpecificCOmpiler

radai-rosenblatt (approved):

PR #317, opened 2022-04-21
Generated class check for spotbugs

li-ukumar:


spotbugs-plugin/src/test/java/com/linkedin/avroutil1/spotbugs/GeneratedTestClass.java
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  @Override
+  public Object get(int field) {
+    return null;
+  }
+
+  @Override
+  public void put(int field, Object value) {
+
```
radai-rosenblatt, line 26: you should add some code here that would normally trigger a violation, like Schema.toString() or something, no?

li-ukumar: Added a method which violates toString schema serialization rule.

radai-rosenblatt (approved):

Issue #319, opened 2022-04-25
extend RandomRecordGenerator to handle self-referencing schemas


radai-rosenblatt:
realistically loops should only happen in combination with unions (otherwise since avro does not allow null values for anything a schema would be infinite in size without a union with null somewhere in the loop).

so it should be enough to keep a stack of types visited during the current record generation and when evaluating a union see if any of the branches are already on the stack, and if so have some % chance to avoid those (this config needs to win over avoidNulls)

PR #321, opened 2022-04-25
add RandomRecordGenerator handling for self-referencing schemas

radai-rosenblatt:
errsolves #319


dg-builder (approved):
LGTM

nit: At some point, would it be worth it to remove the `if (popped != of)` checks? Seems like at some point we should have confidence in this code

radai-rosenblatt:
i'll remove those assertions if/when they become a performance problem :-)

PR #324, opened 2022-04-30
An option to stop deserializing record when all read fields are done

idanz:
In cases where the read schema is much smaller than the write schema, it could have huge performance benefits to just processing the record when the deserializer already got all the fields required by the read schema.
this option is false by default and needs to be turned on via a system property AVRO_FAST_DESERIALIZER_BREAK_EARLY

### DIFF
--- a/avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_10/FastGenericDeserializerGeneratorTest_shouldBeAbleToBreakEarly_GenericDeserializer_8273243328577643558_404527301978058684.java
+++ b/avro-fastserde/src/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_10/FastGenericDeserializerGeneratorTest_shouldBeAbleToBreakEarly_GenericDeserializer_8273243328577643558_404527301978058684.java
@@ -1,0 +1,62 @@
+
+package com.linkedin.avro.fastserde.generated.deserialization.AVRO_1_10;
+
+import java.io.IOException;
+import com.linkedin.avro.fastserde.FastDeserializer;
+import org.apache.avro.Schema;
+import org.apache.avro.generic.IndexedRecord;
+import org.apache.avro.io.Decoder;
+import org.apache.avro.util.Utf8;
+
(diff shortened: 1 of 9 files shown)


radai-rosenblatt:
1. this being controlled via sysprop makes it a global all-or-nothing. is there a finer-grained control mechanism available? we have use cases where multiple different libraries in the same JVM might be using this and this could cause issues.
2. if you are reading multiple payloads from the same input stream, this would break, no? since you wont be skipping over the "suffix" of the payload you'd be leaving the input stream somewhere in the middle of a record and would never be able to read the next record out of the stream?
3. wouldnt the same thing be doable by "doctoring" the writer schema? let field F be the last field in the writer schema that is "of interest" - create a "writer prefix schema" that has all fields up to and including F, but nothing beyond, and use that as the writer schema argument?

idanz:
ok you won me over with the last point, let me try that approach and if it checks out (sounds like it should) then we can close this.
As for the other points raised, I haven't found an easy way to propagate this from the top without sacrificing the compatibility with the standard avro deserializer, so some global flag was needed

FelixGV:
This is interesting and would be great to have a workable solution for, but it seems unfortunately tricky to get right...

Radai's point 2 is an important use case: a given input stream could contain multiple payloads in a row, and we need to maintain a correct behavior in those cases.

It might be the case that implementing this safely requires an interplay between a specialized input stream that "remembers" the tail end of the short-circuited record just in case you end up reading one more record after that. Essentially, some kind of safe laziness, rather than outright omission of the deser. That kind of coupling makes things difficult though.

The schema doctoring approach probably works but may not look ultra clean either. I don't know what is the idea approach. Please keep us posted on your progress! Would love to hear back.

Thanks.

idanz:
I tried the approach and it works well enough for me, the biggest annoyance is that I need it in multiple places and now it's not "out of the box", however I don't have a good fix for the issues you raised other than saying that for me a global flag is good enough. But I understand your reluctance.
I think keeping state of the previous record would be very hard, more so if you consider the next record might not even use the same deserializer. so for my case I will close this and use this workaround, thank you!

PR #325, opened 2022-05-02
Add avro-builder module for building schema files

majisourav99:
Mostly a port of LinkedIn internal avro-schema-builder to be used in Venice where only used CLI options are --input/--output.


FelixGV:
Is there an overlap between this and: https://github.com/linkedin/avro-util/tree/master/avro-codegen/src/main/java/com/linkedin/avro/codegen

IIRC, the above codegen module doesn't fully work, and should therefore have no users... perhaps its content should be substituted by that of the current PR?

@radai-rosenblatt can you chime in?

radai-rosenblatt:
package should be com.linkedin.avroutil1 (this is to allow for future major refactors to have a different package name and co-exist with older code like jackson1/2)

radai-rosenblatt:
some classes still under the "old" namespace - src/main/java/com/linkedin/avro/builder/AvroSchemaBuilderUtils.java ?

PR #333, opened 2022-05-17
make compiler detection more robust, also delay complaints about compiler to point-of-use

radai-rosenblatt:
also dump jcenter for maven central in light of recent build issues

### DIFF
--- a/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
+++ b/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
@@ -49,17 +49,27 @@ import org.apache.avro.specific.SpecificRecord;
 public class AvroCompatibilityHelper {
 
   private static final AvroVersion DETECTED_VERSION;
-  private static final AvroVersion DETECTED_COMPILER_VERSION;
+  private static volatile AvroVersion DETECTED_COMPILER_VERSION;
+  private static volatile Exception COMPILER_DETECTION_ISSUE;
   private static final AvroAdapter ADAPTER;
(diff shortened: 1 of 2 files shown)


helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
```diff
@@ -925,33 +954,25 @@ private static AvroVersion detectAvroCompilerVersion() {
...
-          velocityEngineGetPropertyMethod.invoke(velocityEngine, "file.resource.loader.class");
-
-      return fileResourceLoaderClassProperty != null;
-
-      // We should never reach these exceptions
-    } catch (Exception expected) {
-      throw new IllegalStateException("unexpected exception while retrieving velocity engine config",
-          expected.fillInStackTrace());
-    }
+  private static boolean checkIfVelocityEngineFileResourceLoaderClassExists(Class<?> specificCompilerClass) throws Exception {
```
dg-builder, line 957: Do we want this to throw an uncaught exception if there is any issue with this method?

dg-builder: Nevermind, I see where the try-catch was moved to

dg-builder (approved):

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
radai-rosenblatt: we catch it in the static initializer

PR #334, opened 2022-05-18
go to old json-p artifacts to avoid downstream collisions

radai-rosenblatt:
fun times with duplicatea on the classpath


dg-builder (approved):
LGTM! Thank you 🙏

PR #338, opened 2022-05-19
Properly resolve external refs with inferred namespaces

dg-builder:
Say you have a schema
```
{
  "type": "record",
  "namespace": "com.acme",
  "name": "TestRecordWithExternalReference",
  "doc": "uses com.acme.TestRecord with no namespace",
  "fields": [
    {
      "name": "testRecordField",
      "type": "TestRecord"
    }
  ]
}
```

`testRecordField` refers to another schema which should both check if `TestRecord` (with the null namespace) exists or check if the record with inherited namespace from `TestRecordWithExternalReference` (`com.acme.TestRecord`) exists. However, currently, `parser` does not handle this properly.


radai-rosenblatt (approved):

parser/src/main/java/com/linkedin/avroutil1/parser/avsc/AvroParseContext.java
```diff
@@ -71,19 +71,37 @@ public void resolveReferences() {
...
+                                + inheritedName + ". Only one should exist.";
+                        System.err.println(msg);
+                    }
+                } else if (simpleNameResolution != null) {
+                    ref.setResolvedTo(simpleNameResolution.getDefinedNamedSchemas().get(simpleName));
+                } else {
                     //fqcn is unresolved in this context
-                    externalRefs.computeIfAbsent(fqcn, s -> new ArrayList<>(1)).addAll(refs);
-                    return;
+                    externalReferences.add(ref);
```
dg-builder, line 102: Is there a use in creating an `unresolvedReferences` object?

parser/src/main/java/com/linkedin/avroutil1/parser/avsc/AvroParseContext.java
```diff
@@ -28,7 +28,7 @@ public class AvroParseContext {
     private boolean sealed = false;
     private Map<String, AvscParseResult> knownNamedSchemas = null;
     private Map<String, List<AvscParseResult>> duplicates = null;
-    private Map<String, List<SchemaOrRef>> externalReferences = null;
+    private List<SchemaOrRef> externalReferences = null;
```
dg-builder, line 31: @radai-rosenblatt `externalReferences` is not used at all. Does it have a purpose or is it something we can clean up?

parser/src/main/java/com/linkedin/avroutil1/model/SchemaOrRef.java
```diff
@@ -22,30 +22,36 @@ public class SchemaOrRef implements LocatedCode {
      * the FQCN of a referenced schema, if this represents a reference
      */
     private final String ref;
+    /**
+     * the namespace of the parent schema, which is necessary for reference resolution
```
radai-rosenblatt, line 26: "which might be necessary for reference resolution if the ref is not a fullname"

parser/src/main/java/com/linkedin/avroutil1/model/SchemaOrRef.java
```diff
@@ -62,6 +68,25 @@ public String getRef() {
         return ref;
     }
 
+    // null if ref already is a FQN or if there is no parent namespace.
+    public String getInheritedName() {
```
radai-rosenblatt, line 72: just like in avro-3512, ".SomeRecord" might be legal avro syntax (and is a fullname reference to SomeRecord in the null namespace).

i think your code returns "" (empty string) for both "SomeRecord" and ".SomeRecord"

parser/src/main/java/com/linkedin/avroutil1/parser/avsc/AvscFileParseContext.java
```diff
@@ -64,9 +64,9 @@ public class AvscFileParseContext {
      */
     protected List<AvscIssue> issues = new ArrayList<>();
     /**
-     * references in this avsc to FQCNs that are not in this avsc (to be resolved by a wider context)
+     * references in this avsc that are not in this avsc (to be resolved by a wider context)
```
radai-rosenblatt, line 67: what is the map key ?

parser/src/main/java/com/linkedin/avroutil1/parser/avsc/AvroParseContext.java
radai-rosenblatt: this is what we will use to lookup any still-unresolved references "elsewhere" - like dependency jars

parser/src/main/java/com/linkedin/avroutil1/parser/avsc/AvroParseContext.java
```diff
@@ -71,19 +71,37 @@ public void resolveReferences() {
...
+                    inheritedNameResolution = knownNamedSchemas.get(inheritedName);
+                }
+
+                // The namespace may be inherited from the parent schema's context or may already be defined in the
+                // name. There may be multiple resolutions for a simple name (either from the null namespace or from
+                // the inherited namespace).
+                if (inheritedNameResolution != null) {
+                    ref.setResolvedTo(inheritedNameResolution.getDefinedNamedSchemas().get(inheritedName));
+                    if (simpleNameResolution != null) {
+                        String msg =
```
radai-rosenblatt, line 93: AvscParseResult has "issues". this should be added to "issues" for the schema with the ref in question

parser/src/main/java/com/linkedin/avroutil1/parser/avsc/AvroParseContext.java
radai-rosenblatt: yes - loking them up i other projects/dependencies

parser/src/main/java/com/linkedin/avroutil1/parser/avsc/AvscFileParseContext.java
```diff
@@ -206,12 +203,28 @@ private void resolveReferences(SchemaOrRef possiblyRef) {
...
+
+            String inheritedName = possiblyRef.getInheritedName();
+            if (inheritedName != null) {
+                inheritedNameResolution = definedNamedSchemas.get(inheritedName);
+            }
+
+            if (inheritedNameResolution != null) {
+                possiblyRef.setResolvedTo(inheritedNameResolution);
+                if (simpleNameResolution != null) {
+                    String msg =
```
radai-rosenblatt, line 219: should be an AvscIssue in "issues"

parser/src/main/java/com/linkedin/avroutil1/parser/avsc/AvroParseContext.java
dg-builder: Currently, there is no ability to add new issues to an existing `AvscParseResult`. Do you see there being any side-effects with modifying the existing `issues` in `AvscParseResult`?

radai-rosenblatt: should be fine to add issues. we dont claim the model is thread-safe and this happens during parsing, so the results have not been "finalized" yet

PR #346, opened 2022-06-15
fix parsing of default values for unions

radai-rosenblatt:


parser/src/main/java/com/linkedin/avroutil1/parser/avsc/AvscParser.java
```diff
@@ -381,8 +381,14 @@ private AvroNamedSchema parseNamedSchema(
                     AvroLiteral defaultValue = null;
                     if (fieldDefaultValueNode != null) {
                         if (fieldSchema.isResolved()) {
-                            LiteralOrIssue defaultValurOrIssue = parseLiteral(fieldDefaultValueNode, fieldSchema.getSchema(), fieldName.getValue(), context);
+                            AvroSchema defaultValueExpectedSchema = fieldSchema.getSchema();
+                            if (defaultValueExpectedSchema.type() == AvroType.UNION) {
+                                //(legal) default values are expected to match the 1st union branch
```
PawanHegde, line 386: Is the other case not allowed, or just not the suggested way?

radai-rosenblatt: it will 100% not work if the default value is not the 1st union type, but avro (depends on version and config) is very likely to only explode at runtime and not parse time

PR #352, opened 2022-06-24
introduce SchemaParseConfiguration.LOOSE_NUMERICS and allow control over validation of numeric default value types

radai-rosenblatt:
this allows configuring validation of numeric default value types separately from the more "coarse" validation of default values.

this allows accepting 0.0 for integers and 0 for floating points.

note that modern avro (>= 1.10) performs this validation as part of default value validation (if enabled)


srramach (approved):
LGTM with a nit.

PR #353, opened 2022-06-30
String field and method representation by config

li-ukumar:
default representation of a string field can be set using the config for field and method's return value between String, Charsequence and Utf-8.

### DIFF
--- a/avro-codegen/src/main/java/com/linkedin/avroutil1/codegen/SpecificRecordClassGenerator.java
+++ b/avro-codegen/src/main/java/com/linkedin/avroutil1/codegen/SpecificRecordClassGenerator.java
@@ -6,10 +6,12 @@
 
 package com.linkedin.avroutil1.codegen;
 
+import com.linkedin.avroutil1.compatibility.AvroRecordUtil;
 import com.linkedin.avroutil1.compatibility.AvroVersion;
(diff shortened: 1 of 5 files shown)


avro-codegen/src/main/java/com/linkedin/avroutil1/codegen/SpecificRecordClassGenerator.java
```diff
@@ -208,24 +209,23 @@ protected JavaFileObject generateSpecificRecord(AvroRecordSchema recordSchema, S
...
-                .build());
-      } else {
-        classBuilder.addField(
-            FieldSpec.builder(CLASSNAME_SPECIFIC_DATA, "MODEL$", Modifier.PRIVATE,
-                Modifier.STATIC)
-                .initializer(CodeBlock.of("$T.get()", CLASSNAME_SPECIFIC_DATA))
-                .build());
-      }
-
+    if(config != null && config.getMinimumSupportedAvroVersion().laterThan(AvroVersion.AVRO_1_7)) {
```
radai-rosenblatt, line 212: i think it will be better to substitute a null config with a default instance, or throw at the top for no config. would save you from a lot of null checks

radai-rosenblatt (approved):

PR #354, opened 2022-07-12
add ConfigurableSchemaComparator class to allow for nuance when testing for schema equality

radai-rosenblatt:
its generally impossible to write "portable" (across avro versions) code that relies on the behavior of Schema.equals(), for the following reasons:

1. avro 1.4 discards non-string properties on types and fields, yet props are part of equals(). this means that schemas that are equals() under 1.4 are not so under more modern avro
2. modern versions of avro do not allow floating point default values for integer fields and vice versa. older avro did not complain about those and it is sometimes required to support "loose" comparison of those for legacy reasons
3. avro Schema.equals() does not take aliases into account yet there are situations where differences in aliases are important
4. similarly to above, we want to be able to compare logical types and java string types

class ConfigurableSchemaComparator is meant to allow solving all these issues and providing a consistent equals() implementation across all supported avro versions

### DIFF
--- a/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvroAdapter.java
+++ b/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvroAdapter.java
@@ -103,7 +103,7 @@ public interface AvroAdapter {
 
   boolean fieldHasDefault(Schema.Field field);
 
-  boolean defaultValuesEqual(Schema.Field a, Schema.Field b, boolean looseNumerics);
+  boolean defaultValuesEqual(Schema.Field a, Schema.Field b, boolean allowLooseNumerics);
 
   Set<String> getFieldAliases(Schema.Field field);
 
@@ -136,12 +136,30 @@ public interface AvroAdapter {
 
   void setFieldPropFromJsonString(Schema.Field field, String propName, String valueAsJsonLiteral, boolean strict);
 
+  /**
+   * compare json properties on 2 schema fields for equality.
+   * @param a a field
+   * @param b another field
+   * @param compareStringProps true to compare string properties (otherwise ignored)
+   * @param compareNonStringProps true to compare all other properties (otherwise ignored). this exists because avro 1.4
+   *                              doesnt handle non-string props at all and we want to be able to match that behaviour under any avro
+   * @return if field properties are equal, under the configuration parameters above
+   */
   boolean sameJsonProperties(Schema.Field a, Schema.Field b, boolean compareStringProps, boolean compareNonStringProps);
 
   String getSchemaPropAsJsonString(Schema schema, String propName);
 
   void setSchemaPropFromJsonString(Schema field, String propName, String valueAsJsonLiteral, boolean strict);
 
+  /**
+   * compare json properties on 2 schemas for equality.
+   * @param a a schema
+   * @param b another schema
+   * @param compareStringProps true to compare string properties (otherwise ignored)
+   * @param compareNonStringProps true to compare all other properties (otherwise ignored). this exists because avro 1.4
+   *                              doesnt handle non-string props at all and we want to be able to match that behaviour under any avro
+   * @return if schema properties are equal, under the configuration parameters above
+   */
   boolean sameJsonProperties(Schema a, Schema b, boolean compareStringProps, boolean compareNonStringProps);
 
   String getEnumDefault(Schema s);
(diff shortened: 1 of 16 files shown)


helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvroAdapter.java
```diff
@@ -134,10 +136,14 @@ default FieldBuilder newFieldBuilder(String name) {
...
   void setFieldPropFromJsonString(Schema.Field field, String propName, String valueAsJsonLiteral, boolean strict);
 
+  boolean sameJsonProperties(Schema.Field a, Schema.Field b, boolean compareStringProps, boolean compareNonStringProps);
+
   String getSchemaPropAsJsonString(Schema schema, String propName);
 
   void setSchemaPropFromJsonString(Schema field, String propName, String valueAsJsonLiteral, boolean strict);
 
+  boolean sameJsonProperties(Schema a, Schema b, boolean compareStringProps, boolean compareNonStringProps);
+
```
dg-builder, line 146: why do we need two methods here?

helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvroAdapter.java
```diff
@@ -134,10 +136,14 @@ default FieldBuilder newFieldBuilder(String name) {
 
   void setFieldPropFromJsonString(Schema.Field field, String propName, String valueAsJsonLiteral, boolean strict);
 
+  boolean sameJsonProperties(Schema.Field a, Schema.Field b, boolean compareStringProps, boolean compareNonStringProps);
```
dg-builder, line 139: Maybe we could document somewhere why we support enabling/disabling checking string props and non-string props?

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/ConfigurableSchemaComparator.java
```diff
@@ -0,0 +1,210 @@
...
+        return true;
+      case BYTES:
+        return true;
+
+      //named types
+
+      case ENUM:
+        return a.getFullName().equals(b.getFullName())
+            && (!considerAliases || hasSameAliases(a, b))
+            && a.getEnumSymbols().equals(b.getEnumSymbols());
```
dg-builder, line 79: What if the enum symbols are ordered differently? Are they still equal?

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/SchemaComparisonConfiguration.java
```diff
@@ -0,0 +1,75 @@
...
+package com.linkedin.avroutil1.compatibility;
+
+
+public class SchemaComparisonConfiguration {
+  /**
+   * behaves like avro &lt;= 1.7.2 - non-string props on fields or types are ignored
+   */
+  public static final SchemaComparisonConfiguration PRE_1_7_3 = new SchemaComparisonConfiguration(
+      true, false, false, false, true, false
+  );
```
dg-builder, line 16: nit:  any way we can use something like named parameters here(i.e. a builder pattern?). otherwise, this isn't that distinguishable from binary 011100001 😛

helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvroAdapter.java
```diff
@@ -103,6 +103,8 @@ BinaryDecoder newBinaryDecoder(byte[] bytes, int offset,
 
   boolean fieldHasDefault(Schema.Field field);
 
+  boolean defaultValuesEqual(Schema.Field a, Schema.Field b, boolean looseNumerics);
```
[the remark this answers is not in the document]
radai-rosenblatt: allowLooseNumerics work for you ?

helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvroAdapter.java
radai-rosenblatt: one for schemas, the other for fields. they dont share a common parent class across all avro versions

helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvroAdapter.java
radai-rosenblatt: will do. answer is because avro 1.4 only treats the string props, and we want to be able to replicate this behaviour under any avro

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/ConfigurableSchemaComparator.java
radai-rosenblatt: no, and i think List.equals() covers us for that

PR #356, opened 2022-07-19
add validation to Helper.parse() for contents after the end of an avsc (see AVRO-3560)

radai-rosenblatt:
avro uses a buffered json parser and ignores anything left in the buffer after reading the schema out,
this means that trailing "junk content" after the end of the schema is tolerated.

this PR adds validation to Helper.parse() to look for such junk

### DIFF
--- a/helper/tests/helper-tests-allavro/src/test/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelperParsingTest.java
+++ b/helper/tests/helper-tests-allavro/src/test/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelperParsingTest.java
@@ -232,4 +232,27 @@ public class AvroCompatibilityHelperParsingTest {
       }
     }
   }
+
+  @Test
+  public void testParseTrailingContent() throws Exception {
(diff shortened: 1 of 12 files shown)


FelixGV:
What will it do if it finds such junk? Fail? Or trim it?

radai-rosenblatt:
if junk is not whitespace an IllegalArgument is thrown, indicating what the junk is and where

helper/tests/helper-tests-allavro/src/test/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelperParsingTest.java
```diff
@@ -232,4 +232,23 @@ public void testParsingHorribleLooseNumericDefaults() throws Exception {
       }
     }
   }
+
+  @Test
+  public void testParseTrailingContent() throws Exception {
```
PawanHegde, line 237: Maybe one more check to assert that the loose config continues to allow dangling content?

PawanHegde (approved):
LGTM

PR #357, opened 2022-07-20
Complete implementation for AVRO_UTIL codegen in SchemaBuilder

dg-builder:


avro-builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/own/AvroUtilCodeGenOp.java
```diff
@@ -52,29 +51,10 @@ public void run(OperationContext opContext) throws Exception {
       return;
     }
     LOGGER.info("found " + avscFiles.size() + " avsc schema files");
+    generator.setInputs(config.getInputRoots());
```
radai-rosenblatt, line 55: i dislike the idea of bumdling dircovery of schema files, parsing of schema files and (in the future) classpath lookup of referenced schemas into generator. i'd rather keep generator to be "just" the code-gen, and be able to handle these separately

radai-rosenblatt: also in the future we'd be running validations on these schemas before generating them.

dg-builder: Agreed. I can unbundle discovery, parsing, and code-gen in another PR. cc @li-ukumar in case you have thoughts here

li-ukumar:
@dg-builder Please update the PR description with what this change aims to achieve.

radai-rosenblatt (approved):

PR #358, opened 2022-07-21
Use SpecificRecordGenerationConfig instead of CodeGenerator

dg-builder:
I did not realize that  `com.linkedin.avro.codegen.CodeGenerator` was not using @li-ukumar's Specific Record code generation. I have reverted most of that change and plugged in `SpecificRecordClassGenerator` instead.

### DIFF
--- a/avro-builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/own/AvroUtilCodeGenOp.java
+++ b/avro-builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/own/AvroUtilCodeGenOp.java
@@ -124,11 +124,9 @@ public class AvroUtilCodeGenOp implements Operation {
       }
 
       try (FileOutputStream fos = new FileOutputStream(outputFile, false);
-          OutputStreamWriter writer = new OutputStreamWriter(fos, StandardCharsets.UTF_8)) {
-        Reader reader = javaClass.openReader(true);
+          OutputStreamWriter writer = new OutputStreamWriter(fos, StandardCharsets.UTF_8);
+          Reader reader = javaClass.openReader(true)) {
         IOUtils.copy(reader, writer);
-        reader.close();
-
         writer.flush();
         fos.flush();
       } catch (Exception e) {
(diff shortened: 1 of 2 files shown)


avro-builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/own/AvroUtilCodeGenOp.java
```diff
@@ -51,10 +69,76 @@ public void run(OperationContext opContext) throws Exception {
...
+          OutputStreamWriter writer = new OutputStreamWriter(fos, StandardCharsets.UTF_8)) {
+        Reader reader = javaClass.openReader(true);
+        IOUtils.copy(reader, writer);
+        reader.close();
+
+        writer.flush();
+        fos.flush();
+      } catch (Exception e) {
+        throw new IllegalStateException("while writing file " + outputFile, e);
+      }
```
dg-builder, line 141: Let me know if there's a cleaner way to handle this Reader -> Writer copy.

avro-builder/src/test/java/com/linkedin/avroutil1/builder/SchemaBuilderTest.java
```diff
@@ -97,6 +97,7 @@ private File locateTestProjectsRoot() {
     if (!file.isDirectory() || !file.canRead()) {
       throw new IllegalStateException("test-projects root folder " + file.getAbsolutePath() + " not a folder or is unreadable");
     }
+    System.err.println(file.getAbsolutePath());
```
li-ukumar, line 100: Is this added for debugging purposes? 
If yes, Add a prefix like "File absolute path:" for context and move from err.println -> out.println if its just info

dg-builder: Yes, it was for debugging purposes. I will remove

avro-builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/own/AvroUtilCodeGenOp.java
radai-rosenblatt: nit - you might be able to move the reader initialization into the try(), which would save you the need to close() it.

radai-rosenblatt (approved):

li-ukumar (approved):

PR #360, opened 2022-08-02
adds support for importable source to avro-util's codegen/builder

dg-builder:
This PR adds support for the `--include` statement that `VanillaProcessedCodeGenOp` already implements.

One detail of note: if there are no `--include` schemas, all `--input` schemas are allowed to import from each other. Is this the functionality we want?

### DIFF
--- a/avro-builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/own/AvroUtilCodeGenOp.java
+++ b/avro-builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/own/AvroUtilCodeGenOp.java
@@ -21,8 +21,6 @@ import java.io.FileOutputStream;
 import java.io.OutputStreamWriter;
 import java.io.Reader;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.Files;
-import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
@@ -59,7 +57,6 @@ public class AvroUtilCodeGenOp implements Operation {
     Set<File> avscFiles = new HashSet<>();
     Set<File> importableFiles = new HashSet<>();
     String[] extensions = new String[]{BuilderConsts.AVSC_EXTENSION};
-
 
     for (File inputRoot : config.getInputRoots()) {
       avscFiles.addAll(FileUtils.listFiles(inputRoot, extensions, true));
--- a/avro-builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/own/AvroUtilCodeGenOp.java
+++ b/avro-builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/own/AvroUtilCodeGenOp.java
@@ -75,10 +75,10 @@ public class AvroUtilCodeGenOp implements Operation {
     LOGGER.info("found " + (avscFiles.size() + nonImportableFiles.size()) + " avsc schema files");
 
     AvroParseContext context = new AvroParseContext();
-    List<AvscParseResult> parsedFiles = new ArrayList<>();
+    Set<AvscParseResult> parsedFiles = new HashSet<>();
 
-    parseAvscFiles(avscFiles, true, context, parsedFiles);
-    parseAvscFiles(nonImportableFiles, false, context, parsedFiles);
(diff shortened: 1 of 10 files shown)


avro-builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/own/AvroUtilCodeGenOp.java
```diff
@@ -46,38 +49,42 @@ public AvroUtilCodeGenOp(CodeGenOpConfig config) {
...
           "unable to create destination folder " + config.getOutputSpecificRecordClassesRoot());
     }
 
-    List<Path> avscFiles = new ArrayList<>();
+    Set<File> avscFiles = new HashSet<>();
+    Set<File> importableFiles = new HashSet<>();
+    String[] extensions = new String[]{BuilderConsts.AVSC_EXTENSION};
+
+
     for (File inputRoot : config.getInputRoots()) {
```
radai-rosenblatt, line 61: i think input should be the importable one, and the non-importable sources should get another name.
long term we want to have all inputs be importable, and eliminate usage of non-importable sources

avro-builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/own/AvroUtilCodeGenOp.java
```diff
@@ -46,38 +49,42 @@ public AvroUtilCodeGenOp(CodeGenOpConfig config) {
...
+    }
+
+    if (config.getIncludeRoots() != null) {
+      for (File includeRoot : config.getIncludeRoots()) {
+        importableFiles.addAll(FileUtils.listFiles(includeRoot, extensions, true));
+      }
     }
-    if (avscFiles.isEmpty()) {
+
+    // If there are no importable files, then all of avscFiles should be marked as importable.
```
radai-rosenblatt, line 71: i think this would result in unintuitive behaviour

dg-builder: Good point, I'll adjust

avro-builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/own/AvroUtilCodeGenOp.java
```diff
@@ -103,6 +104,20 @@ public void run(OperationContext opContext) throws Exception {
 
   }
 
+  private void parseAvscFiles(Set<File> avscFiles, boolean areFilesImportable, AvroParseContext context,
```
radai-rosenblatt, line 107: javadoc - what is parsedFiles? (i assume files previously parsed during the same "run" which are importable to files being parsed now)

avro-builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/own/AvroUtilCodeGenOp.java
```diff
@@ -46,38 +47,38 @@ public AvroUtilCodeGenOp(CodeGenOpConfig config) {
...
-      AvscParseResult fileParseResult = parser.parse(p);
-      Throwable parseError = fileParseResult.getParseError();
-      if (parseError != null) {
-        throw new IllegalArgumentException("failed to parse file " + p.toAbsolutePath(), parseError);
-      }
-      context.add(fileParseResult);
-      parsedFiles.add(fileParseResult);
-    }
+    parseAvscFiles(avscFiles, true, context, parsedFiles);
+    parseAvscFiles(nonImportableFiles, false, context, parsedFiles);
```
radai-rosenblatt, line 81: now that we have multiple "parse runs" - do we need to check for duplicate definitions across them ? (adding a TODO would suffice for now)

radai-rosenblatt (approved):

avro-builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/own/AvroUtilCodeGenOp.java
dg-builder: Good point. There's a TODO in this file still so I'll keep it there for now.

PR #363, opened 2022-08-05
implement delayed parsing of field default values

radai-rosenblatt:


parser/src/test/resources/schemas/TestRecordWithEnumField.avsc
```diff
@@ -0,0 +1,13 @@
+{
+  "type": "record",
+  "namespace": "com.acme",
+  "name": "TestRecordWithEnumField",
+  "doc": "nothing fancy",
+  "fields": [
+    {
+      "name": "enumField",
+      "type": "TestEnum",
+      "default": "SYMBOL_B"
```
FelixGV, line 10: So the purpose of this change is to allow references to externally defined types, rather than needing to inline everything?

radai-rosenblatt: allowing avsc reuse has always been the goal. this PR is mostly about completing missing functionality and fixing bugs

radai-rosenblatt:
Yes. standard avsc does not allow for any reuse of schemas. We need the
ability to reuse schemas from other files

PR #364, opened 2022-08-08
[builder] separate plugin context from operation context

dg-builder:
* separates operation context from plugin context
* Adds `Set<File> avroFiles` to operation context
* Not sure if we should add `Set<AvroSchema> avroSchemas` since they won't necessarily be generated if we use vanilla codegen

### DIFF
--- a/avro-builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/OperationContext.java
+++ b/avro-builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/OperationContext.java
@@ -6,35 +6,60 @@
 
 package com.linkedin.avroutil1.builder.operations.codegen;
 
+import com.linkedin.avroutil1.builder.operations.AvroSchemaUtils;
+import com.linkedin.avroutil1.model.AvroSchema;
 import java.io.File;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Set;
+import org.apache.avro.Schema;
 
 
 /**
  * common context to all {@link com.linkedin.avroutil1.builder.operations.Operation}s run during an execution
  */
 public class OperationContext {
-  private Set<File> avroFiles;
   private final HashMap<String, Object> context;
+  private Set<File> avroFiles;
+  private Set<AvroSchema> avroSchemas;
 
   public OperationContext() {
     this.context = new HashMap<>();
   }
 
-  public void setAvroFiles(Set<File> avroFiles) {
-    if (this.avroFiles != null) {
+  public void addParsedSchemas(Set<AvroSchema> avroSchemas, Set<File> avroFiles) {
+    if (this.avroFiles != null || this.avroSchemas != null) {
       throw new IllegalStateException("Cannot initialize avro files twice");
     }
-    this.avroFiles = avroFiles;
+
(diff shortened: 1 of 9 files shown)


avro-builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/OperationContext.java
```diff
@@ -0,0 +1,53 @@
...
+import java.io.File;
+import java.util.HashMap;
+import java.util.Set;
+
+
+/**
+ * common context to all {@link com.linkedin.avroutil1.builder.operations.Operation}s run during an execution
+ */
+public class OperationContext {
+  private Set<File> avroFiles;
```
radai-rosenblatt, line 18: we want the parsed AvroSchemas available as well, likely even more important than the "raw" input files.

avro-builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/own/AvroUtilCodeGenOp.java
```diff
@@ -100,6 +100,10 @@ public void run(OperationContext opContext) throws Exception {
 
     writeJavaFilesToDisk(specificRecords, config.getOutputSpecificRecordClassesRoot());
 
+    Set<File> allAvroFiles = new HashSet<>(avscFiles);
+    allAvroFiles.addAll(nonImportableFiles);
+    opContext.setAvroFiles(allAvroFiles);
```
radai-rosenblatt, line 105: make the collection unmodifiable for extra safety? dont give plugin authors the ability to assume this is mutable (in a meaningful way)

dg-builder: How do you recommend doing that? I see Guava has an `ImmutableSet` class but I was wondering if you had a better way in mind

avro-builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/OperationContext.java
dg-builder: As I mentioned in the description, adding avroSchemas will work if we use our own codegen but not if we do vanilla `org.apache.avro` codegen. How do you want to handle the case where we do Vanilla avro codegen? 

A strawman approach would be to add both `apacheAvroSchemas` and `avroUtilSchemas` to the context and plugins should expect only one of them to be defined. The issue then is that plugins have to handle both cases.

radai-rosenblatt: our own model should be considered part of the builder API.
there should also be utility code to convert from our own AvroSchema to vanilla org.apache.avro.Schema instances

avro-builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/own/AvroUtilCodeGenOp.java
radai-rosenblatt: Collections.unmodifiable...() family of methods. built into t he JDK

radai-rosenblatt (approved):

PR #369, opened 2022-08-14
minor cleanups around code-gen

radai-rosenblatt:
- support for writing out default values for map and union fields
- print timings in code-gen op


avro-builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/own/AvroUtilCodeGenOp.java
```diff
@@ -84,31 +88,61 @@ public void run(OperationContext opContext) throws Exception {
     //resolve any references across files that are part of this op (anything left would be external)
     context.resolveReferences();
 
+    long parseEnd = System.currentTimeMillis();
+
+    Map<String, AvscParseResult> allNamedSchemas = context.getAllNamedSchemas();
+    Map<String, List<AvscParseResult>> duplicates = context.getDuplicates();
+
+    LOGGER.info("parsed {} named schemas in {} millis, {} of which have duplicates",
+        allNamedSchemas.size(), parseEnd - scanEnd, duplicates.size());
```
dg-builder, line 97: In the avro-schemas builder, there are multiple options on what to do when duplicates are found. do we want to support the same config options?

radai-rosenblatt: yes, but in a later PR. broadly the options should be "fail-on-any-dup", "fail-when-meaningfully-different" and "log a warning and skulk in the corner"

PR #374, opened 2022-08-24
Inner schemas  in record generated as separate files

li-ukumar:


avro-builder/builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/own/AvroUtilCodeGenOp.java
```diff
@@ -114,11 +114,11 @@ public void run(OperationContext opContext) throws Exception {
       AvroNamedSchema namedSchema = fileParseResult.getDefinedSchema(fullname);
 
       try {
-        JavaFileObject javaFileObject = generator.generateSpecificRecordClass(
+        List<JavaFileObject> javaFileObjects = generator.generateSpecificClass(
```
radai-rosenblatt, line 117: i think the generator should be a "simple" pirce of code - 1 named type in, 1 java class out.
the work of recursing into inner types and generating them should be higher-order logic.

PR #387, opened 2022-09-20
Enforces using Avro 702 Handling & adds min avro version config

dg-builder:
Enforces using Avro 702 Handling in 2 ways
1. Changes the `--enableAvro702Handling` default value from `false` -> `true`
2. Logs a warning if the user attempts to set the value to `true`

Eventually, we will deprecate this field, but for now we can track who still tries to disable 702 handling.

This PR also adds handling of the `--minAvroVersion` to avro-util's custom codegen.


avro-builder/builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/own/AvroUtilCodeGenOp.java
```diff
@@ -58,6 +58,10 @@ public void run(OperationContext opContext) throws Exception {
           "unable to create destination folder " + config.getOutputSpecificRecordClassesRoot());
     }
 
+    if (!config.isAvro702Handling()) {
```
radai-rosenblatt, line 61: overly strict (sadly). can we print a warning instead?

PR #391, opened 2022-09-23
adds import from classpath functionality to AvroUtilCodeGenOp

dg-builder:
supports `--includeClasspath` flag. When resolving references, unresolved schemas will be searched for on the classpath. Some changes were needed to codegen setup in order to not generate `.java` specific records for classpath files.

One option, if deemed fit, would be to move the import from classpath functionality into `AvroParseContext`. However, since this functionality is not part of org.apache.avro, it is probably best to keep it separate.

unit tests added as well

### DIFF
--- a/avro-builder/builder/src/main/java/com/linkedin/avroutil1/builder/SchemaBuilder.java
+++ b/avro-builder/builder/src/main/java/com/linkedin/avroutil1/builder/SchemaBuilder.java
@@ -19,7 +19,6 @@ import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Iterator;
 import java.util.List;
 import java.util.ServiceLoader;
 import java.util.Set;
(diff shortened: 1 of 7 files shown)


config/checkstyle/suppressions.xml
```diff
@@ -6,4 +6,5 @@
 
 <suppressions>
   <suppress checks=".*" files="[\\/]build[\\/]generated[\\/].*" />
-</suppressions>
\ No newline at end of file
+  <suppress checks=".*" files="[\\/]test[\\/]java[\\/]simpleproject[\\/]SimpleRecord.java" />
```
li-ukumar, line 9: Why do we need this?

dg-builder: `SimpleRecord` is generated code so it’s causing a handful of checkstyle issues. We need to include the generated class in the classpath in order to test if imports from classpath will work.

li-ukumar (approved):

PR #393, opened 2022-09-28
avro-builder tests using own codegen

li-ukumar:
tests in avro-builder moved to using own codegen. Currently only 1.4


avro-builder/tests/codegen-14/build.gradle
```diff
@@ -34,7 +34,8 @@ task runOwnCodegen {
       main = 'com.linkedin.avroutil1.builder.SchemaBuilder'
       args = [
           "--input", "$projectDir/src/main/avro",
-          "--output", "$buildDir/generated/sources/avro/java/main"
+          "--output", "$buildDir/generated/sources/avro/java/main",
+          "--generator", "AVRO_UTIL"
```
dg-builder, line 38: why change just avro 1.4 codegen to use `--generator=AVRO_UTIL`? what about the other versions?

avro-codegen/src/main/java/com/linkedin/avroutil1/codegen/SpecificRecordClassGenerator.java
```diff
@@ -499,14 +505,26 @@ private void populateBuilderClassBuilder(TypeSpec.Builder recordBuilder, AvroRec
...
-              .beginControlFlow("catch (org.apache.avro.AvroMissingFieldException  e)")
-              .addStatement("e.addParentField(record.getSchema().getField($S))", escapedFieldName)
-              .addStatement("throw e")
+              .endControlFlow();
+
+
+
+          if (config.getMinimumSupportedAvroVersion().laterThan(AvroVersion.AVRO_1_8)) {
+            buildMethodCodeBlockBuilder.beginControlFlow("catch (org.apache.avro.AvroMissingFieldException e)")
+                .addStatement("com.linkedin.avroutil1.compatibility.exception.AvroUtilMissingFieldException avroUtilException = new com.linkedin.avroutil1.compatibility.exception.AvroUtilMissingFieldException(e)")
```
dg-builder, line 518: Are these `Exception`s the first cases where our custom codegen will rely on `avro-util` as a dependency?

avro-builder/tests/codegen-14/build.gradle
li-ukumar: We only have avro1.4 in the test module for now

avro-codegen/src/main/java/com/linkedin/avroutil1/codegen/SpecificRecordClassGenerator.java
li-ukumar: codegen already uses Compatibility helpers, so this will not result in a new dependency. 
Avro uses different exception classes depending on their versions which changes what our generated code looks like

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/exception/AvroUtilMissingFieldException.java
```diff
@@ -0,0 +1,39 @@
...
+
+package com.linkedin.avroutil1.compatibility.exception;
+
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.avro.AvroRuntimeException;
+import org.apache.avro.Schema;
+
+
+public class AvroUtilMissingFieldException extends AvroUtilException{
```
dg-builder, line 15: Could we add to the javadoc as to how this exception is used? I think it would be useful for future readers (including ourselves :P ) when we are trying to figure out how it is used

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/StringConverterUtil.java
```diff
@@ -0,0 +1,35 @@
...
+ * See License in the project root for license information.
+ */
+
+package com.linkedin.avroutil1.compatibility;
+
+import java.nio.charset.StandardCharsets;
+import org.apache.avro.util.Utf8;
+
+
+public class StringConverterUtil {
```
dg-builder, line 13: Where is this used?

radai-rosenblatt: i believe the intent is to call this from generated code if/when converting between Utf8 fields and String getters/setters is needed ?

avro-builder/builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/own/AvroUtilCodeGenOp.java
```diff
@@ -202,8 +202,9 @@ public void run(OperationContext opContext) throws Exception {
           "unresolved referenced to external schemas: " + context.getExternalReferences());
     }
 
+
     SpecificRecordClassGenerator generator = new SpecificRecordClassGenerator();
-    List<JavaFileObject> specificRecords = new ArrayList<>(allNamedSchemas.size());
+    HashSet<JavaFileObject> specificRecords = new HashSet<>(allNamedSchemas.size());
```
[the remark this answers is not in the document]
li-ukumar: Converted to a Set to handle duplicate internal types, which we are now generating. Its a choice between extra effort to de-dup here vs overwriting the previous one during writeOnDisk

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/StringConverterUtil.java
li-ukumar: This is being written inside generated code for conversion between Charsequence types (String, CharSequence and Utf8)
Converting it to a Util with static methods in the next commit

dg-builder: Ah I see. Then similar to the comment above, it could be good to add this note to the javadoc or as a normal comment

dg-builder (approved):
🚢

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/StringConverterUtil.java
```diff
@@ -0,0 +1,40 @@
...
+ * String util used in codegen to convert between Charsequence types (String, Utf8, Charsequence)
+ */
+public class StringConverterUtil {
+
+  private StringConverterUtil() {
+  }
+
+  private static void validateArgs(Object arg) {
+    if(! (arg instanceof CharSequence)) {
+      throw new IllegalArgumentException("StringConverterUtil accepts only CharSequence arguments");
```
radai-rosenblatt, line 22: nit - would be better to also print the bad value: 
```
String.valueOf(arg) + " (a " + arg.getClass().name() + ") is not a CharSequence"
```
(handle null)

radai-rosenblatt (approved):

PR #404, opened 2022-10-19
Store utf8 runtime, String accessors

li-ukumar:
- public Charsequence fields store Utf8s during runtime to support 1.4 functionality.

- type String Accessors 

**Tests:**
- Updated test case to verify getter setter and public field
- tracking build successful with this snapshot


li-ukumar:
generates: 

`
  public java.lang.CharSequence stringField;
  
  public java.lang.String getStringField() {
    return this.stringField == null ? null : String.valueOf(this.stringField);
  }

  public void setStringField(java.lang.String stringField) {
    if (this.stringField == null || this.stringField instanceof org.apache.avro.util.Utf8) {
      this.stringField = stringField;
    }
    else {
      this.stringField = new org.apache.avro.util.Utf8(stringField.toString());
    }
  }
`

radai-rosenblatt (approved):

Issue #407, opened 2022-10-25
Fastavro code generator generates classes with String literals of length > 65K


pugupta-linkedin:
After a new schema evolution, fastavro code generator created a class which uses a string literal as arguments to an exception class. The string literal has lenght greater than 65K. The java compiler cannot compile such classes

dg-builder:
fixed in #408

PR #408, opened 2022-10-26
[avro-util] break down large string literals into chunks

pugupta-linkedin:
Fix for https://github.com/linkedin/avro-util/issues/407

We notice that the code gen creates Java classes which have huge string literals.
This change prevents that.

For large literals it breaks them down into chunks and uses StringBuilder.append().append()... calls on the smaller literals.

Has unit test


avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -633,11 +635,20 @@ private void processUnion(JVar unionSchemaVar, final String name, final Schema u
 
         if (null == readerOptionSchema) {
           // This is the same exception that vanilla Avro would throw in this circumstance
-          thenBlock._throw(JExpr._new(codeModel.ref(AvroTypeException.class)).arg(
-              JExpr.lit("Found " + optionSchema + ", expecting " + (readerSchemaNotAUnion ? unionReaderSchema.toString()
-                      : unionReaderSchema.getTypes().toString()))));
+          String fullExceptionString = "Found " + optionSchema + ", expecting " + (readerSchemaNotAUnion ? unionReaderSchema.toString()
+              : unionReaderSchema.getTypes().toString());
+
+          if (fullExceptionString.length() <= MAX_LENGTH_OF_STRING_LITERAL) {
```
gaojieliu, line 641: Can we fold this condition check into the newly created function: `composeLargeStringLiteral`? and rename it to be `composeStringLiteral` since it can handle both small and large strings.

avro-fastserde/src/test/java/com/linkedin/avro/fastserde/FastSpecificDeserializerGeneratorTest.java
```diff
@@ -824,6 +824,23 @@ public void shouldReadUnionOfRecordFieldsWithSameEnumFieldName(Boolean whetherUs
     Assert.assertEquals(getField((MyRecordV2) getField(deserRecord, "unionField"), "enumField"), MyEnumV2.VAL3);
   }
 
+  @Test(groups = {"deserializationTest"})
```
gaojieliu, line 827: This new test should generate some test classes when executing the test case, so can you also check in the generated test classes?

Normally, we will divide the commit into two:
1. Main code change including test cases.
2. Generated codes.

An example:
https://github.com/linkedin/avro-util/pull/326/commits

gaojieliu (changes requested):

avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
pugupta-linkedin: Done

gaojieliu (approved):
Thanks a lot!

PR #409, opened 2022-10-31
Updated handling of Collections of Charseq and Obj

li-ukumar:
- ListTransformer won't transform List of non-charseq objects
- Unions containing charseq (or List/Map of) to be covered by their collection Transformers


avro-codegen/src/main/java/com/linkedin/avroutil1/codegen/SpecificRecordClassGenerator.java
```diff
@@ -51,7 +51,7 @@
 import java.util.stream.Collectors;
 import javax.lang.model.element.Modifier;
 import javax.tools.JavaFileObject;
-
+import org.apache.logging.log4j.util.Strings;
```
radai-rosenblatt, line 54: is this used? what for?

radai-rosenblatt: is it only for Strings.EMPTY?

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/collectiontransformer/ListTransformer.java
```diff
@@ -14,38 +14,67 @@
...
+        } else if (item instanceof CharSequence) {
+          hasCharSeq = true;
+          ret.add(StringConverterUtil.getString(item));
+        } else {
+          ret.add(item);
+        }
       }
+      return hasCharSeq? Collections.unmodifiableList(ret) : list;
+    } else {
+      throw new UnsupportedOperationException("Supports only Lists");
```
radai-rosenblatt, line 77: you should print the type of the input here to make errors easier to resolve

PR #413, opened 2022-11-04
add test showing string type conversion as part of generic to specific conversion

radai-rosenblatt:


li-ukumar (approved):
Exciting stuff!

PR #416, opened 2022-11-16
Codegen builder update : Stop assuming field type Records have builders

li-ukumar:
- Builder classes don't assume fields with record types follow Builder pattern.
- doesn't produce 'fieldName'+builder field inside Builder class
- Sets value for such fields directly ( without using their builders)

Builder -> build() method moves from
```       if (minBuilder != null) {
          try {
            record.min = this.minBuilder.build();
          }
          catch (org.apache.avro.AvroRuntimeException e) {
            com.linkedin.avroutil1.compatibility.exception.AvroUtilMissingFieldException avroUtilException = new com.linkedin.avroutil1.compatibility.exception.AvroUtilMissingFieldException(e);
            avroUtilException.addParentField(record.getSchema().getField("min"));
            throw avroUtilException;
          }
        }
        else {
          record.min = fieldSetFlags()[6] ? this.min : (vs111.Amount) com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper.getSpecificDefaultValue(fields()[6]);
        }
```

to 
```
record.min = fieldSetFlags()[6] ? this.min : (vs111.Amount) com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper.getSpecificDefaultValue(fields()[6]);


radai-rosenblatt (approved):
as a baseline implementation not assuming dependency classes have builders is the safest approach.

longer-term though i'd like to make this a class-by-class basis - examine any dependency/external classes, see if they have builders, and generate code accordingly.

PR #417, opened 2022-11-22
Change to override equals in AvroSchema

li-ukumar:
- Adds overridden equals method in AvroSchema.
- Adds overridden equals method for AvroLiterals


helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroCompatibilityHelper.java
```diff
@@ -108,13 +103,6 @@ public class AvroCompatibilityHelper {
     }
   }
 
-  /**
-   * returns the detected runtime version of avro, or null if none found
-   * @return the version of avro detected on the runtime classpath, or null if no avro found
-   */
-  public static AvroVersion getRuntimeAvroVersion() {
```
radai-rosenblatt, line 115: you should keep this, add @Deprecated, and make it call the new location (for backwards compatiblity reasons)

radai-rosenblatt (approved):

PR #423, opened 2022-11-29
Codegen: Charseq + string constructors for all string fields Record

li-ukumar:
If all fields in a record are of string type, generate 2 all arg constructors
- With all string args
- with all Charseq args


radai-rosenblatt (approved):

radai-rosenblatt:
nit - i think the extra constructor (with CharSequence args) needs to be deprecated

PR #424, opened 2022-11-29
add AvroRecordUtil.setStringField() for simple string fields, validate avro default string behaviours

radai-rosenblatt:
handling collections of strings TBD still. 
also no caching for performance yet


helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroRecordUtil.java
```diff
@@ -230,6 +231,69 @@ public static GenericRecord specificRecordToGenericRecord(
...
+      //generic record
+      //TODO - make this honor schema logical types or runtime avro version?
+      record.put(field.pos(), value);
+      return record;
+    }
+    Class<? extends IndexedRecord> recordClass = record.getClass();
+    //look for public field 1st
+    Field[] classFields = recordClass.getFields();
+    for (Field classField : classFields) {
+      //TODO - take into account avro name mangling?
```
li-ukumar, line 250: +1. method used in Codegen can be made common.

li-ukumar (approved):

li-ukumar (approved):

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroRecordUtil.java
radai-rosenblatt: will be done as part of a follow-up

PR #428, opened 2022-12-14
field setting support for more complex generated class properties

radai-rosenblatt:


helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvroSchemaUtil.java
```diff
@@ -529,6 +529,35 @@ public static Schema resolveUnionBranchOf(Object value, Schema union) {
...
+   * Utf8 vs java.lang.Strings
+   * @param schema a schema
+   * @return true if value under schema could possibly involve strings
+   */
+  public static boolean schemaContainsString(Schema schema) {
+    if (schema == null) {
+      return false;
+    }
+    boolean hasString = false;
+    switch (schema.getType()) {
```
dg-builder, line 544: Curious if we should handle records here? If not, let's include it in the description/name given that this is a public util

helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvroSchemaUtil.java
```diff
@@ -529,6 +529,35 @@ public static Schema resolveUnionBranchOf(Object value, Schema union) {
...
+      case STRING:
+      case MAP: //map keys are always strings, regardless of values
+        return true;
+      case UNION:
```
xy-builder, line 548: nit: we not need to iterate all union members, we can return true when seeing the first true on schemaContainsString(branch).

helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvroSchemaUtil.java
radai-rosenblatt: this doesnt traverse into records. updated the javadoc

radai-rosenblatt: also it would hurt its intended usage if it traversed into records. when setting fields we do not need to convert records (any record field strings would have been taken care of before)

PR #432, opened 2022-12-21
Updated codegen builder initializer and Avrorecord deepcopy

li-ukumar:
- AvroRecordUtil deepcopy handles Union and Charseq type
- newBuilder expected string type is same as field type used in builder.

### DIFF
--- a/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroRecordUtil.java
+++ b/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroRecordUtil.java
@@ -711,6 +711,11 @@ public class AvroRecordUtil {
           outputMap.put(outputKey, outputElement);
         }
         return outputMap;
+
+      case UNION:
+        inputValueActualSchema = AvroSchemaUtil.resolveUnionBranchOf(inputValue, inputSchema);
+        outputValueActualSchema = AvroSchemaUtil.resolveUnionBranchOf(inputValue, outputSchema);
+        return deepConvert(inputValue, inputValueActualSchema, outputValueActualSchema, context, stringRepresentation);
     }
     String inputClassName = inputValue == null ? "null" : inputValue.getClass().getName();
     throw new UnsupportedOperationException("dont know how to convert " + inputType + " " + inputValue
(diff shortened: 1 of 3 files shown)


helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroRecordUtil.java
```diff
@@ -711,6 +711,11 @@ protected static Object deepConvert(
           outputMap.put(outputKey, outputElement);
         }
         return outputMap;
+
+      case UNION:
+        inputValueActualSchema = AvroSchemaUtil.resolveUnionBranchOf(inputValue, inputSchema);
```
radai-rosenblatt, line 716: this switch is on the OUTPUT schema type. is it possible that in case of schema evolution the input schema is not a union?

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroRecordUtil.java
```diff
@@ -776,6 +781,7 @@ private static CharSequence toString(CharSequence inputStr, StringRepresentation
     switch (desired) {
       case String:
         return String.valueOf(inputStr);
+      case CharSequence:
```
radai-rosenblatt, line 784: nit - add a comment that this falls-through on purpose

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroRecordUtil.java
li-ukumar: Possibly, if input and output schemas are different. Which would result in an exception.

radai-rosenblatt: can we add a TODO here? or try this in a unit test?

radai-rosenblatt (approved):

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroRecordUtil.java
li-ukumar: update: newBuilder(otherBuilder) -> deepCopy flow won't allow another schema version so this shouldn't be possible there is another flow which uses this?

radai-rosenblatt (approved):

PR #434, opened 2023-01-05
update generated code post-processing for WRITER$ to avoid deadlock in 1.11

radai-rosenblatt:
so, turns out
```
WRITER$ = new SpecificDatumWriter(SCHEMA$); 
```
creates a deadlock under avro 1.11, since as part of the ctr call it will call SpecificData.getForClass(), which will attempt to re-enter CHM.compute() in SpecificData.classCache.

the solution is to provide the instance of SpecificData to use
```
WRITER$ = new SpecificDatumWriter(SCHEMA$, MODEL$);
```
which is what modern avro does out of the box (supporting logical types in avro is via a per-class instance of SpecificData stored in MODEL$).

since the relevant ctr isnt available before avro ~1.7.4 i added a helper method to access it, making the resulting generated code be
```
WRITER$ = AvroCompatibilityHelper.newSpecificDatumWriter(SCHEMA$, MODEL$)
```

### DIFF
--- a/helper/tests/codegen-111/src/main/compat-avro/under111/DeeplyNested.avsc
+++ b/helper/tests/codegen-111/src/main/compat-avro/under111/DeeplyNested.avsc
@@ -1,0 +1,29 @@
+{
+  "type": "record",
+  "name": "Outer",
+  "namespace": "under111",
+  "fields": [
+    {
+      "name": "outerRef",
+      "type": {
+        "type": "record",
+        "name": "Middle",
+        "fields": [
+          {
+            "name": "middleRef",
+            "type": {
+              "type": "record",
+              "name": "Inner",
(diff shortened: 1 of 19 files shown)


helper/tests/codegen-111/src/main/compat-avro/under111/DeeplyNested.avsc
```diff
@@ -0,0 +1,29 @@
+{
```
dg-builder, line 1: is this schema used at all?

radai-rosenblatt: yes. it generates classes Outer Middle and Inner later used in the test (to deserialize into)

PR #436, opened 2023-01-05
update avro 1.11 to 1.11.1. update some tests to reflect avro 1.11.1 breakage

radai-rosenblatt:
looks like 1.11.1 no longer validates namespaces across writer and reader schemas?


dg-builder (approved):
🚢

PR #437, opened 2023-01-06
apply avro-1.11 codegen fixes to our own codegen

radai-rosenblatt:
same thing as was done in https://github.com/linkedin/avro-util/pull/434

### DIFF
--- a/avro-codegen/src/main/java/com/linkedin/avroutil1/codegen/SpecificRecordClassGenerator.java
+++ b/avro-codegen/src/main/java/com/linkedin/avroutil1/codegen/SpecificRecordClassGenerator.java
@@ -32,8 +32,10 @@ import com.squareup.javapoet.FieldSpec;
 import com.squareup.javapoet.JavaFile;
 import com.squareup.javapoet.MethodSpec;
 import com.squareup.javapoet.ParameterSpec;
+import com.squareup.javapoet.ParameterizedTypeName;
 import com.squareup.javapoet.TypeName;
 import com.squareup.javapoet.TypeSpec;
+import com.squareup.javapoet.TypeVariableName;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


avro-codegen/src/main/java/com/linkedin/avroutil1/codegen/SpecificRecordClassGenerator.java
```diff
@@ -391,46 +392,60 @@ protected JavaFileObject generateSpecificRecord(AvroRecordSchema recordSchema, S
...
+        .methodBuilder("writeExternal")
+        .addException(IOException.class)
+        .addParameter(java.io.ObjectOutput.class, "out")
+        .addModifiers(Modifier.PUBLIC)
+        .addCode(CodeBlock
+            .builder()
+            .addStatement("$L.write(this, com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper.newBinaryEncoder(out))", "WRITER$")
+            .build());
+    //no @Overrides annotation since under old avro the parent class is not externalizable
+    //TODO - add this is min avro version is high enough
```
dg-builder, line 442: what do you mean by this TODO? in general, curious why we are punting on this (is it difficult / low value, is it not possible right now, etc

dg-builder (approved):
overall, lgtm! just curious about the TODO but not a blocker :)

avro-codegen/src/main/java/com/linkedin/avroutil1/codegen/SpecificRecordClassGenerator.java
radai-rosenblatt: the overrides annotation has no value. we strip it out to avoid compilation issues under old avro.
we need to spend a more significant amount of time figuring out what "min target avro version" should actually mean.

PR #438, opened 2023-01-09
make testing code-gen tasks declare inputs and outputs for incremental build support

radai-rosenblatt:
this drops the 2nd+ run time for ./gradlew :helper:tests:helper-tests-allavro:test down from ~50 seconds to ~8 seconds on my machine


li-ukumar (approved):
LGTM

PR #439, opened 2023-01-12
Improve Jacoco code coverage

xiaoyu-yang-gh:
This is to improve the Jacoco code coverage:
(1) change the generate html report to make it more readable.
(2) include more subprojects to the report.
(3) exclude the generated code from the report.
(4) add unit tests for some zero covered tests.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -273,6 +273,13 @@ task codeCoverageReport(type: JacocoReport) {
           //add source sets from all relevant submodules
           sourceSets subproject.sourceSets.main
         }
+        classDirectories.setFrom(files(classDirectories.files.collect {
+          fileTree(dir: it, exclude: [
+              "by*/*",
+              "under*/*",
+              "vs*/*"
+          ])
+        }))
       }
     }
     executionData.setFrom(fileTree(project.rootDir.absolutePath).include("**/build/jacoco/*.exec"))
@@ -281,7 +288,8 @@ task codeCoverageReport(type: JacocoReport) {
   reports {
     xml.enabled true
     xml.destination = new File("${buildDir}/reports/jacoco/report.xml")
-    html.enabled false //useful as a debug aid and for local dev
+    html.enabled true
+    html.destination = new File("${buildDir}/reports/jacoco/html")
     csv.enabled false
   }
 }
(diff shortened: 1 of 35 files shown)


build.gradle
```diff
@@ -273,6 +273,13 @@ task codeCoverageReport(type: JacocoReport) {
           //add source sets from all relevant submodules
           sourceSets subproject.sourceSets.main
         }
+        classDirectories.setFrom(files(classDirectories.files.collect {
+          fileTree(dir: it, exclude: [
```
radai-rosenblatt, line 277: nit - can you add a comment explaining this is done to explode some classes from coverage by package?

common/build.gradle
```diff
@@ -0,0 +1,14 @@
...
+ * Licensed under the BSD 2-Clause License (the "License").
+ * See License in the project root for license information.
+ */
+
+plugins {
+    id "java-library"
+    id "checkstyle"
+}
+
+dependencies {
```
radai-rosenblatt, line 12: nit - i think you can remove the entire dependencies {} section if there's nothing there ?

common/src/main/java/com/linkedin/avroutil1/common/ExcludeFromJacocoGeneratedReport.java
```diff
@@ -0,0 +1,24 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+
+/**
+ * Annotation that can be used to exclude code from jacoco coverage.
+ */
+
+@Retention(RetentionPolicy.RUNTIME)
```
radai-rosenblatt, line 20: can we reduce this to SOURCE or something?

radai-rosenblatt: no, we cant

radai-rosenblatt (approved):

common/build.gradle
xiaoyu-yang-gh: removed

build.gradle
xiaoyu-yang-gh: added

PR #447, opened 2023-02-03
Fix issue to create specific type of object, instead of GenericDataRe…

xiaoyu-yang-gh:
When using the generated builder to copy an object, throw exception:
org.apache.avro.generic.GenericData$Record cannot be cast to com.linkedin.events.EventHeader
java.lang.ClassCastException: org.apache.avro.generic.GenericData$Record cannot be cast to com.linkedin.events.EventHeader

Fix this issue to create the specific type of object, instead of GenericData$Record during deep copying the object.

### DIFF
--- a/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/CompatibleSpecificRecordBuilderBase.java
+++ b/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/CompatibleSpecificRecordBuilderBase.java
@@ -86,7 +86,7 @@ public class CompatibleSpecificRecordBuilderBase {
     return false;
   }
 
-
+  @SuppressWarnings("unchecked")
   protected  <T> T deepCopyField (Object inputValue, Schema fieldSchema,
       String stringRepresentationValue) {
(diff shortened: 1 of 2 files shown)


helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/CompatibleSpecificRecordBuilderBase.java
```diff
@@ -86,7 +86,7 @@ protected static boolean isValidValue(Schema.Field f, Object value) {
     return false;
   }
 
-
+  @SuppressWarnings("unchecked")
   protected  <T> T deepCopyField (Object inputValue, Schema fieldSchema,
```
li-ukumar, line 90: Please add the test case which produces this error as well.

li-ukumar (changes requested):

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/CompatibleSpecificRecordBuilderBase.java
xiaoyu-yang-gh: added test

radai-rosenblatt (approved):

PR #448, opened 2023-02-03
Builder: Support codegen on avsc files from a jar/zip file

dg-builder:


avro-builder/builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/util/AvscFileFinderUtil.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Enumeration;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.jar.JarFile;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
+import org.apache.commons.io.FileUtils;
+
+
+public class AvscFileFinderUtil {
```
radai-rosenblatt, line 25: javadoc on the class at least would be nice

avro-builder/builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/util/AvscFileFinderUtil.java
```diff
@@ -0,0 +1,74 @@
...
+  private static final String[] EXTENSIONS = new String[]{BuilderConsts.AVSC_EXTENSION};
+  private static final String TEMP_DIR_NAME = "tempCompressedAvroSrc";
+
+  private AvscFileFinderUtil() {
+  }
+
+  public static Collection<File> findFiles(File inputRoot) throws IOException {
```
radai-rosenblatt, line 32: worth documenting. the fact that this will fish inside zips and jars anywhere under the given input root is not immediately obvious and intuitive

radai-rosenblatt: also - what would this do with dupes? same file (contents) inside and outside of a zip?

dg-builder: I'll document but this "fishing" behavior should be as expected: if you pass in a jar/zip by name as an input, I would expect that it looks inside that jar/zip. 

What this method won't do: recurse through your input root and look for jars/zips and get all avsc files inside of those zips/jars.

dg-builder: Dupe handling happens after this. avscs found in zips should be treated as any other avsc and how it is handled will depend on your configured dupe behavior

radai-rosenblatt (approved):

PR #451, opened 2023-02-27
[codegen] reference class fields using 'this.'

dg-builder:
In the generated `get(int field)` method, if the field name was `field`, the local method argument was referenced, not the class field. Prefixing thos lookups with `this.` fixes this issue.

Alternatively, we could consider postfixing all method arguments with `$` or in some way that there would never be a conflict (I imagine this is why `SCHEMA$` is used).

Before
Screen Shot 2023-02-26 at 7 14 32 PM

After
Screen Shot 2023-02-26 at 7 18 08 PM


radai-rosenblatt (approved):
proper use of "this." is significantly better then implicitly relying on obscure naming conventions

PR #459, opened 2023-03-24
Remove additional character check in deserializeAsGeneric for json

xiaoyu-yang-gh:
The additional characters check in deserializeAsGeneric does not work as expected.
(1) sometimes it does not throw even additional garbage characters appended to json.
(2) json parser does not read the last `}` character, so the `}` might be left in the InputStream, and be treated as garbage character.


xiaoyu-yang-gh:
discard, will create a new one.

PR #469, opened 2023-04-04
Updating build file to explain the need of custom JDK11 tasks

alex-dubrouski:


radai-rosenblatt (approved):
thank you for doing this

PR #471, opened 2023-04-07
Adding a fix for VH read in cacheFromByteBuffer method

alex-dubrouski:


radai-rosenblatt (commented):
would it be possible to add a unit test to verify this?

alex-dubrouski:
> would it be possible to add a unit test to verify this?

Working on it right now

fastserde/avro-fastserde/src/main/java11/com/linkedin/avro/fastserde/BufferBackedPrimitiveFloatList.java
```diff
@@ -290,7 +290,7 @@ private void cacheFromByteBuffer() {
       if (!isCached) {
         setupElements(this, this.size);
         for (int i = 0; i < size; i++) {
-          elements[i] = (float)VH.get(byteBuffer, i);
+          elements[i] = (float)VH.get(byteBuffer, i * FLOAT_SIZE);
```
FelixGV, line 293: Why not do `i += FLOAT_SIZE` instead of `i++` in the for loop, rather than multiplying on each iteration? Maybe JIT would already do that since it's equivalent but anyhow...

FelixGV (approved):
Thanks for the fix! Left a minor comment, which I can give you another ship it for if you choose to address it.

fastserde/avro-fastserde/src/main/java11/com/linkedin/avro/fastserde/BufferBackedPrimitiveFloatList.java
radai-rosenblatt: you'd need to also update the condition to be i < size * FLOAT_SIZE to match. and hypothetically this might overflow for large lists .... oh, and loops on longs are slower.

so needs benchmarking if you really want to try this

FelixGV: Loops on longs are slower?

FelixGV (approved):

fastserde/avro-fastserde/src/main/java11/com/linkedin/avro/fastserde/BufferBackedPrimitiveFloatList.java
majisourav99: BTW do we need the caching ? I added it to help for bytebuffer access, with UnSafe access we might not help much. Good idea to benchmark it again.

radai-rosenblatt: YMMV, but at least this - https://developers.redhat.com/articles/2022/08/25/optimize-loops-long-variables-java#
also i remember long loops dont stop at safe points which means no GC while the loop is running, which could mean longer GC pause times (i remember reading this but couldnt find a source)

PR #473, opened 2023-04-07
upgrade to gradle 7

radai-rosenblatt:
working towards gradle 8, but 7 is a good milestone


dg-builder (approved):

fastserde/avro-fastserde/build.gradle
```diff
@@ -49,7 +49,7 @@ dependencies {
   implementation "org.apache.commons:commons-lang3:3.4"
   implementation "com.sun.codemodel:codemodel:2.6"
 
-  compile ("org.apache.avro:avro:1.10.2") {
+  compileOnly ("org.apache.avro:avro:1.10.2") {
```
alex-dubrouski, line 52: I wonder if we can use "java11CompileOnly" for jdk11 classes Avro dependency and remove custom pom logic. But need to test whether declared dependency on Avro will re-appear in POM file. If not - this will reduce amount of build customizations

radai-rosenblatt: yes, compileOnly should avoid exporting the dependency, which means it shouldnt be in the pom.xml

alex-dubrouski: Could you please run publishToMavenLocal and verify that resulting pom file does not have dependency on Avro? Sorry, just want to make sure we do not resurface this problem. Thanks

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.8.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.6.1-all.zip
```
alex-dubrouski, line 3: JFYI: this also allows to run build on JDK17+, so later on we can modify CI recipe to do that

radai-rosenblatt: yes, but that is not at all representative of the "typical" JVM this code is expected to run on?

fastserde/avro-fastserde/build.gradle
radai-rosenblatt: i have, and there is no avro dependency in the pom file for fast-serde

gradle/wrapper/gradle-wrapper.properties
alex-dubrouski: Yeah, all of the clients are on JDK8 or 11 I assume, that's why I said it paves the way to build/test on JDK17 in future, but no obligations to implement it.

fastserde/avro-fastserde/build.gradle
alex-dubrouski: Thanks

alex-dubrouski (approved):

PR #476, opened 2023-04-14
Nested arrays cause 'item-end when a 'x' was expected" error

asautins:
Schemas with nested arrays generate the following error when deserialized records with fastserde:

```
Attempt to process a item-end when a string was expected.
```

This PR adds a test case (`FastNestedArrayTest.java`) to validate behavior


fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGeneratorBase.java
```diff
@@ -85,6 +86,16 @@ protected ListIterator<Symbol> actionIterator(FieldAction action) {
     while (actionIterator.hasNext()) {
       Symbol symbol = actionIterator.next();
 
+      if (Symbol.Kind.REPEATER.equals(symbol.kind)
+          && "array-end"
+          .equals(
+              getSymbolPrintName(
```
radai-rosenblatt, line 92: this method is fairly slow, and this class (Symbol) is back ported - so we can change it.

is it worth making getPrintName() public so it could be accessed here?

asautins: Terminal symbol toString returns the print name.  Terminal is the only Symbol with printName.  Re-worked to use Terminal method toString to get printName

fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGeneratorBase.java
```diff
@@ -85,6 +85,14 @@ protected ListIterator<Symbol> actionIterator(FieldAction action) {
     while (actionIterator.hasNext()) {
       Symbol symbol = actionIterator.next();
 
+      if (Symbol.Kind.REPEATER.equals(symbol.kind)  &&
```
gaojieliu, line 88: Please leave some javadoc to help other developers understand the full context.

gaojieliu (approved):

PR #477, opened 2023-04-18
Implementing util method to compare 2 avro schemas and return list of…

ajinkya-dande-git:
### Description:
Implementing `findDifference` method in `ConfigurableAvroSchemaComparator` which can be used to compare two Avro schema and return a `List<AvroSchemaDifference>` containing a collection of exact location, type, and summary of differences in schemaA, schemaB (the schema being compared). This can be achieved by updating the existing `equals` implementation to update the `List<AvroSchemaDifference>` instead of returning a boolean. 

**Code Changes:**
- Added `findDifference` method in `ConfigurableAvroSchemaComparator`
- Updated the `equals` method to update the differences instead of returning a boolean 
- Minor bug fixes: 
    - Record schema additional field checks
    - Union branches reference check
- Unit testing the util

**As a reviewer, please help with looking for potential issues for :**
 - Testing that has been done and more testing that can be done 
 - Coding conventions, code reuse and sound logic
 - Performance and scalability

### Testing Done
- New Tests: Added new avsc schemas `UtilTester1` and `UtilTester2` which contained multiple differences. Added unit test to findDifference between these avsc's. 
- Existing Tests: Existing tests passed. 
- Manual Tests: Debugging the changes and fix bugs.

### DIFF
--- a/parser/src/main/java/com/linkedin/avroutil1/model/AvroSchemaDifference.java
+++ b/parser/src/main/java/com/linkedin/avroutil1/model/AvroSchemaDifference.java
@@ -76,7 +76,7 @@ public class AvroSchemaDifference {
    */
   @Override
   public String toString() {
-    return "Type: " + avroSchemaDifferenceType.toString() +
+    return "[" + avroSchemaDifferenceType.toString() + "] " +
         ", SchemaALocation: " + ((schemaALocation != null) ? schemaALocation.toString() : "null") +
         ", SchemaBLocation: " + ((schemaBLocation != null) ? schemaBLocation.toString() : "null") +
         ", DifferenceSummary: " + differenceSummary;
--- a/parser/src/main/java/com/linkedin/avroutil1/model/AvroSchemaDifference.java
+++ b/parser/src/main/java/com/linkedin/avroutil1/model/AvroSchemaDifference.java
@@ -1,0 +1,78 @@
+package com.linkedin.avroutil1.model;
+
+/**
+ * The AvroSchemaDifference class is used to record the differences identified while comparing two Avro schemas. This class
+ * records the code location of the difference identified in schemaA and the corrosponding location in schemaB and records the
+ * summary of difference
+ *  * It supports comparing schemas defined in Avro IDL (.avdl) or Avro schema (.avsc) formats.
+ */
+public class AvroSchemaDifference {
+
+  /**
+   * {@link CodeLocation} of the difference in schemaA
+   */
+  private final CodeLocation schemaALocation;
+
+  /**
+   * {@link CodeLocation} of the difference in schemaB
+   */
+  private final CodeLocation schemaBLocation;
+
+  /**
+   * Enum representing Avro schema difference types.
+   */
+  AvroSchemaDifferenceType avroSchemaDifferenceType;
+
+  /**
+   * Summary of the difference between the schemas
+   */
+  private final String differenceSummary;
+
+  /**
+   * Constructor for creating an AvroSchemaDifference object.
+   *
+   * @param schemaALocation           The {@link CodeLocation} of the difference in schemaA
+   * @param schemaBLocation           The {@link CodeLocation} of the difference in schemaB
+   * @param avroSchemaDifferenceType  The AvroSchemaDifferenceType representing the type of difference
+   * @param differenceSummary         The summary of the difference between the schemas
+   */
+  public AvroSchemaDifference(CodeLocation schemaALocation,
+                              CodeLocation schemaBLocation,
+                              AvroSchemaDifferenceType avroSchemaDifferenceType,
+                              String differenceSummary) {
+    this.schemaALocation = schemaALocation;
+    this.schemaBLocation = schemaBLocation;
+    this.avroSchemaDifferenceType = avroSchemaDifferenceType;
+    this.differenceSummary = differenceSummary;
+  }
+
+  public CodeLocation getSchemaALocation() {
+    return schemaALocation;
+  }
+
+  public CodeLocation getSchemaBLocation() {
+    return schemaBLocation;
+  }
+
+  public AvroSchemaDifferenceType getAvroSchemaDifferenceType() {
+    return avroSchemaDifferenceType;
+  }
+
+  public String getDifferenceSummary() {
+    return differenceSummary;
+  }
+
+  /**
+   * Overrides the default toString() method to return a custom string representation
+   * of the AvroSchemaDifference object.
+   *
+   * @return A string representation of the AvroSchemaDifference object
+   */
+  @Override
+  public String toString() {
+    return "Type: " + avroSchemaDifferenceType.toString() +
+        ", SchemaALocation: " + ((schemaALocation != null) ? schemaALocation.toString() : "null") +
+        ", SchemaBLocation: " + ((schemaBLocation != null) ? schemaBLocation.toString() : "null") +
+        ", DifferenceSummary: " + differenceSummary;
+  }
+}
--- a/parser/src/main/java/com/linkedin/avroutil1/model/AvroSchemaDifference.java
+++ b/parser/src/main/java/com/linkedin/avroutil1/model/AvroSchemaDifference.java
@@ -1,3 +1,9 @@
+/*
+ * Copyright 2022 LinkedIn Corp.
+ * Licensed under the BSD 2-Clause License (the "License").
+ * See License in the project root for license information.
+ */
+
 package com.linkedin.avroutil1.model;
 
 /**
--- a/parser/src/main/java/com/linkedin/avroutil1/model/AvroSchemaDifferenceType.java
+++ b/parser/src/main/java/com/linkedin/avroutil1/model/AvroSchemaDifferenceType.java
@@ -72,11 +72,6 @@ public enum AvroSchemaDifferenceType {
   RECORD_NAME_MISMATCH,
 
   /**
-   * Record field count mismatch between schema A and schema B.
-   */
-  RECORD_FIELD_COUNT_MISMATCH,
-
-  /**
    * Record field name mismatch between schema A and schema B.
    */
   RECORD_FIELD_NAME_MISMATCH,
@@ -87,9 +82,9 @@ public enum AvroSchemaDifferenceType {
   RECORD_DEFAULT_VALUE_MISMATCH,
 
   /**
-   * Default value missing in schema A or schema B.
+   * Default value mismatch in schema A or schema B.
    */
-  DEFAULT_VALUE_MISSING,
+  DEFAULT_VALUE_MISMATCH,
 
   /**
    * Record field position mismatch between schema A and schema B.
--- a/parser/src/main/java/com/linkedin/avroutil1/model/AvroSchemaDifferenceType.java
+++ b/parser/src/main/java/com/linkedin/avroutil1/model/AvroSchemaDifferenceType.java
@@ -99,11 +99,6 @@ public enum AvroSchemaDifferenceType {
   /**
    * Additional field found in schema A or schema B.
    */
-  ADDITIONAL_FIELD,
-
-  /**
-   * Unknown difference type.
-   */
-  UNKNOWN
+  ADDITIONAL_FIELD
 
 }
--- a/parser/src/main/java/com/linkedin/avroutil1/model/AvroSchemaDifferenceType.java
+++ b/parser/src/main/java/com/linkedin/avroutil1/model/AvroSchemaDifferenceType.java
@@ -1,3 +1,9 @@
+/*
+ * Copyright 2022 LinkedIn Corp.
+ * Licensed under the BSD 2-Clause License (the "License").
+ * See License in the project root for license information.
+ */
+
 package com.linkedin.avroutil1.model;
 
 /**
--- a/parser/src/main/java/com/linkedin/avroutil1/model/AvroSchemaDifferenceType.java
+++ b/parser/src/main/java/com/linkedin/avroutil1/model/AvroSchemaDifferenceType.java
@@ -1,0 +1,103 @@
+package com.linkedin.avroutil1.model;
+
+/**
+ * Enum representing Avro schema difference types.
+ */
+public enum AvroSchemaDifferenceType {
+
+  /**
+   * Null schema provided.
+   */
+  NULL_SCHEMA,
+
+  /**
+   * Schema reference mismatch between schema A and schema B.
+   */
+  SCHEMA_REFERENCE_MISMATCH,
+
+  /**
+   * Type mismatch between schema A and schema B.
+   */
+  TYPE_MISMATCH,
+
+  /**
+   * Aliases mismatch between schema A and schema B.
+   */
+  ALIASES_MISMATCH,
+
+  /**
+   * JSON property count mismatch between schema A and schema B.
+   */
+  JSON_PROPERTY_COUNT_MISMATCH,
+
+  /**
+   * JSON property mismatch between schema A and schema B.
+   */
+  JSON_PROPERTY_MISMATCH,
+
+  /**
+   * Enum name mismatch between schema A and schema B.
+   */
+  ENUM_NAME_MISMATCH,
+
+  /**
+   * Enum symbol mismatch between schema A and schema B.
+   */
+  ENUM_SYMBOL_MISMATCH,
+
+  /**
+   * Fixed name mismatch between schema A and schema B.
+   */
+  FIXED_NAME_MISMATCH,
+
+  /**
+   * Fixed size mismatch between schema A and schema B.
+   */
+  FIXED_SIZE_MISMATCH,
+
+  /**
+   * Union size mismatch between schema A and schema B.
+   */
+  UNION_SIZE_MISMATCH,
+
+  /**
+   * Record name mismatch between schema A and schema B.
+   */
+  RECORD_NAME_MISMATCH,
+
+  /**
+   * Record field count mismatch between schema A and schema B.
+   */
+  RECORD_FIELD_COUNT_MISMATCH,
+
+  /**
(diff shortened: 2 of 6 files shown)


parser/src/main/java/com/linkedin/avroutil1/util/ConfigurableAvroSchemaComparator.java
```diff
@@ -84,115 +102,232 @@ private static boolean equals(AvroSchema a, AvroSchema b, SchemaComparisonConfig
...
+        }
+
+        // Compare symbols of enum fields in schemaA and schemaB
+        if (!enumA.getSymbols().equals(enumB.getSymbols())) {
+          AvroSchemaDifference difference = new AvroSchemaDifference(a.getCodeLocation(), b.getCodeLocation(), AvroSchemaDifferenceType.ENUM_SYMBOL_MISMATCH,
+              String.format("Symbols %s of enum field %s in schemaA does not match with symbols %s of enum field %s in schemaB",
+                            enumA.getSymbols().toString(), enumFullNameA,
+                            enumB.getSymbols().toString(), enumFullNameB));
+          differences.add(difference);
+        }
```
xiaoyu-yang-gh, line 137: we can move these code to a method to make it more clean.
same for comparing other types.

parser/src/main/java/com/linkedin/avroutil1/util/ConfigurableAvroSchemaComparator.java
```diff
@@ -39,13 +42,27 @@ public class ConfigurableAvroSchemaComparator {
...
+    validateConfig(config);
+    Set<SeenPair> seen = new HashSet<>(3);
+    List<AvroSchemaDifference> differences = new ArrayList<>();
+    try {
+      equals(a, b, config, seen, differences);
+    } catch (IOException ignored) {
+      // ignored
+    }
+    return differences;
+  }
```
xiaoyu-yang-gh, line 64: you added a new public method **findDifference**, how do you expect users call these two public methods, **equals** and  **findDifference**?
if users call call **equals** first, if returns false, then call **findDifference** to get the difference? this is low efficient, since they compare the schemas twice.

Users call **findDifference** standalone?  the problem is that the exception is not handled in this method. So if IOException is thrown during the method call, this method ignore the exception, the difference returned might be empty, so users get empty differences, and they think two schemas are identical.

Also add Java doc for all public methods.

parser/src/test/java/com/linkedin/avroutil1/util/ConfigurableAvroSchemaComparatorTest.java
```diff
@@ -22,23 +27,89 @@ public class ConfigurableAvroSchemaComparatorTest {
...
+    AvroRecordSchema recordSchema2 = (AvroRecordSchema) validateAndGetAvroRecordSchema(path2);
+    Assert.assertTrue(ConfigurableAvroSchemaComparator.equals(recordSchema1, recordSchema2, SchemaComparisonConfiguration.PRE_1_7_3));
+  }
+
+  @Test
+  public void testNotEquals() throws IOException {
+    AvroRecordSchema schema1 = (AvroRecordSchema) validateAndGetAvroRecordSchema("schemas/UtilTester1.avsc");
+    AvroRecordSchema schema2 = (AvroRecordSchema) validateAndGetAvroRecordSchema("schemas/UtilTester2.avsc");
+    Assert.assertFalse(ConfigurableAvroSchemaComparator.equals(schema1, schema2, SchemaComparisonConfiguration.PRE_1_7_3));
+  }
```
xiaoyu-yang-gh, line 54: why not use dataProvider on all of these test methods? these methods are very similar, so you can just have one method, but use dataProvider to provide different inputs to the test method.

xiaoyu-yang-gh (changes requested):

parser/src/main/java/com/linkedin/avroutil1/util/ConfigurableAvroSchemaComparator.java
xiaoyu-yang-gh, line 64: Also there are code duplicate between method **equals** and **findDifference**.

You can have method **equals** call method **findDifference**:

xiaoyu-yang-gh (changes requested):

parser/src/main/java/com/linkedin/avroutil1/util/ConfigurableAvroSchemaComparator.java
```diff
@@ -39,13 +42,27 @@ public class ConfigurableAvroSchemaComparator {
   public static boolean equals(AvroSchema a, AvroSchema b, SchemaComparisonConfiguration config) {
     validateConfig(config);
     Set<SeenPair> seen = new HashSet<>(3);
+    List<AvroSchemaDifference> differences = new ArrayList<>();
     try {
-      return equals(a, b, config, seen);
+      equals(a, b, config, seen, differences);
+      return differences.isEmpty();
```
dg-builder, line 48: Let's have `equals` call the old internal version of `private static boolean equals(a, b, config, seen)`. The old version failed fast, which is desirable for the `equals` use case.

In particular, Schema Registry uses this implementation to fail fast when checking for schema equivalence and using the changed implementation will be slower (since it finds all differences instead of failing fast)

parser/src/main/java/com/linkedin/avroutil1/model/AvroSchemaDifference.java
```diff
@@ -0,0 +1,84 @@
...
+
+  /**
+   * Overrides the default toString() method to return a custom string representation
+   * of the AvroSchemaDifference object.
+   *
+   * @return A string representation of the AvroSchemaDifference object
+   */
+  @Override
+  public String toString() {
+    return "Type: " + avroSchemaDifferenceType.toString() +
```
dg-builder, line 79:
```suggestion
    return "[" + avroSchemaDifferenceType.toString() + "] " + 
```

parser/src/main/java/com/linkedin/avroutil1/model/AvroSchemaDifferenceType.java
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   * Additional field found in schema A or schema B.
+   */
+  ADDITIONAL_FIELD,
+
+  /**
+   * Unknown difference type.
+   */
+  UNKNOWN
```
dg-builder, line 107: Do we ever hit this case? If not, let's remove :)

parser/src/main/java/com/linkedin/avroutil1/util/ConfigurableAvroSchemaComparator.java
dg-builder: So my suggested implementation is to leave the old `private static boolean equals(a, b, config, seen)` as is and create a new internal method called `findDifferencesEquals` and have `findDifferences` call that helper method.

It would also be useful to add a line or 2 of documentation as to why we chose this implementation and added a bit of duplicated code :)

parser/src/main/java/com/linkedin/avroutil1/util/ConfigurableAvroSchemaComparator.java
dg-builder: I think that `findDifference` and `equals` serve 2 different use cases. `equals` is just if you need to know if the schemas are equivalent, `findDifference` is for user-facing tools to find differences.

parser/src/main/java/com/linkedin/avroutil1/util/ConfigurableAvroSchemaComparator.java
dg-builder: `findEnumDifferences()`

parser/src/main/java/com/linkedin/avroutil1/util/ConfigurableAvroSchemaComparator.java
```diff
@@ -84,115 +102,232 @@ private static boolean equals(AvroSchema a, AvroSchema b, SchemaComparisonConfig
...
+        if (!(!considerAliases || hasSameAliases((AvroFixedSchema) a, (AvroFixedSchema) b))) {
+          AvroSchemaDifference difference = new AvroSchemaDifference(a.getCodeLocation(), b.getCodeLocation(), AvroSchemaDifferenceType.ALIASES_MISMATCH,
+              String.format("Aliases for fixed field %s in schemaA does not match with the aliases for fixed field %s in schemaB",
+                            fixedFullNameA, fixedFullNameB));
+          differences.add(difference);
+        }
 
+        // Compare size of fixed fields in schemaA and schemaB
+        if ((fixedA.getSize() != fixedB.getSize())) {
+          AvroSchemaDifference difference = new AvroSchemaDifference(a.getCodeLocation(), b.getCodeLocation(), AvroSchemaDifferenceType.FIXED_SIZE_MISMATCH,
```
dg-builder, line 164: Same as @xiaoyu-yang-gh's comment above: `findFixedDifferences()`

dg-builder (commented):
overall looks great! 

Main comment is that we shouldn't modify the current `equals` implementation as some internal users rely on its fail-fast performance. Instead, let's make a separate helper method for the findDifferences equality checks and document the reason for somewhat duplicate code in the comments.

parser/src/main/java/com/linkedin/avroutil1/util/ConfigurableAvroSchemaComparator.java
ajinkya-dande-git: Good callout, updated this in the recent commit.

parser/src/main/java/com/linkedin/avroutil1/util/ConfigurableAvroSchemaComparator.java
ajinkya-dande-git: @dg-builder @xiaoyu-yang-gh 
1. Instead of writing separate methods having similar code, I have introduced `fastFail` flag. This flag can be set to `true` for equals and it will fail on the first mismatch. We can set it to `false` for the `findDifference` and the code will go on to find all the differences in the schemas. Plus, this will improve code reusability and optimize performance. 
2. Throwing a `IllegalStateException` for any `IOExceptions`

parser/src/test/java/com/linkedin/avroutil1/util/ConfigurableAvroSchemaComparatorTest.java
ajinkya-dande-git: Thanks for the callout, added the change in the most recent commit

parser/src/main/java/com/linkedin/avroutil1/util/ConfigurableAvroSchemaComparator.java
ajinkya-dande-git: Added a helper method which will be called by both `equals` and `findDifference` methods.

parser/src/main/java/com/linkedin/avroutil1/util/ConfigurableAvroSchemaComparator.java
ajinkya-dande-git: Instead of re-adding `equals` method, added a flag`fastFail` to return on the first identified difference.

parser/src/main/java/com/linkedin/avroutil1/model/AvroSchemaDifference.java
ajinkya-dande-git: Added this code change in the most recent commit.

parser/src/main/java/com/linkedin/avroutil1/model/AvroSchemaDifferenceType.java
ajinkya-dande-git: Done.

parser/src/main/java/com/linkedin/avroutil1/util/ConfigurableAvroSchemaComparator.java
```diff
@@ -39,30 +39,45 @@
  */
 public class ConfigurableAvroSchemaComparator {
 
+  /**
+   * Compares two AvroSchema objects for equality using the specified SchemaComparisonConfiguration. This method is high
+   * performant and return on the first found mismatch.
```
dg-builder, line 44: > and return on the first found mismatch.

it returns a boolean so that is false, correct?

parser/src/main/java/com/linkedin/avroutil1/util/ConfigurableAvroSchemaComparator.java
```diff
@@ -84,115 +121,272 @@ private static boolean equals(AvroSchema a, AvroSchema b, SchemaComparisonConfig
...
+          AvroSchemaDifference difference = new AvroSchemaDifference(aField.getCodeLocation(), bField.getCodeLocation(),
+              AvroSchemaDifferenceType.ALIASES_MISMATCH,
+              String.format("Aliases for %s in schemaA does not match with the aliases for %s in schemaB", a, b));
+          differences.add(difference);
+          if (fastFail) return;
         }
       }
-      return true;
+      // If there is size mismatch between the schemas, we add the additional fields in the differences
+      for (int i = numberOfCommonFields; i < aFields.size(); i++) {
```
dg-builder, line 374: Seems like we are adding the same error twice: we will never have a `RECORD_FIELD_COUNT_MISMATCH` difference without an `ADDITIONAL_FIELD` difference.

I'm thinking we should remove `RECORD_FIELD_COUNT_MISMATCH`?

parser/src/main/java/com/linkedin/avroutil1/util/ConfigurableAvroSchemaComparator.java
```diff
@@ -36,16 +39,45 @@
...
+   * @param a The first {@link AvroSchema} to compare.
+   * @param b The second {@link AvroSchema} to compare.
+   * @param config The {@link SchemaComparisonConfiguration} that specifies the configuration for the comparison.
+   * @return List of {@link AvroSchemaDifference} objects representing the differences between the two AvroSchema objects
+   */
+  public static List<AvroSchemaDifference> findDifference(AvroSchema a, AvroSchema b, SchemaComparisonConfiguration config) {
+    return equals(a, b, config, false);
+  }
+
+  private static List<AvroSchemaDifference> equals(AvroSchema a, AvroSchema b, SchemaComparisonConfiguration config, boolean fastFail) {
```
dg-builder, line 68: I'm thinking that this API could be used in the future where the user just wants to find the first error. Given the current structure, we could now add an optional `failFast` parameter.

We can probably implement this as follows: create a `findDifferences()` and `findDifferencesFailFast()` methods.

dg-builder: This can be a follow-up PR to unblock you

parser/src/main/java/com/linkedin/avroutil1/util/ConfigurableAvroSchemaComparator.java
```diff
@@ -84,115 +121,272 @@ private static boolean equals(AvroSchema a, AvroSchema b, SchemaComparisonConfig
...
+                    aField.getDefaultValue(), aField.getName(), a,
+                    bField.getDefaultValue(), bField.getName(), b));
+            differences.add(difference);
+            if (fastFail) return;
           }
         } else if (aField.hasDefaultValue() || bField.hasDefaultValue()) {
-          //means one field has a default value and the other does not
-          return false;
+          AvroSchemaDifference difference = new AvroSchemaDifference(aField.getCodeLocation(), bField.getCodeLocation(),
+              AvroSchemaDifferenceType.DEFAULT_VALUE_MISSING,
```
dg-builder, line 344: for consistency

```suggestion
              AvroSchemaDifferenceType.DEFAULT_VALUE_MISMATCH,
```

dg-builder (approved):
a few comments but in general, LGTM!

xiaoyu-yang-gh (approved):

parser/src/main/java/com/linkedin/avroutil1/util/ConfigurableAvroSchemaComparator.java
ajinkya-dande-git: That is correct, I also have a unit test which tests and confirms this behavior.

dg-builder: Ok cool, let's update that comment then :)

parser/src/main/java/com/linkedin/avroutil1/util/ConfigurableAvroSchemaComparator.java
ajinkya-dande-git: Its not the same error, `RECORD_FIELD_COUNT_MISMATCH` refers to the number of fields differed in both schemas, whereas `ADDITIONAL_FIELD` refers to the exact field name which is absent from the differing schema.

parser/src/main/java/com/linkedin/avroutil1/util/ConfigurableAvroSchemaComparator.java
ajinkya-dande-git: Is there any scenario where user will need only the first diff? Considered that there is such scenario, user can get all the differences and check the first one :) thoughts?

dg-builder: My thought is that this could end up being used by performance-focused use cases that still want to show a descriptive user message. They can find if the schemas are equal and if not, they can show a single error message

dg-builder: Not urgent though and we can add in the future if needed so feel free to skip.

PR #489, opened 2023-05-10
Legacy schema writer option

li-ukumar:
## What
Adding a legacy option in Avsc writer to generate Avsc before normalization was introduced.
Only canonicalization configs to use non-legacy option.

## Why
Some use cases Hash the serialized schemas and are now incompatible with the normalized changes. While the long-term viability of some of these usecases are questionable, in short term it means normalization will break their systems.

## Testing
Updated Unit tests

snapshot on
- `fdp-utils` : https://lva1-app63607.corp.linkedin.com/s/rnisgewbdaj4e
- `avro-schemas` : https://lva1-app63607.corp.linkedin.com/s/ooz7ieoavr3gc

### DIFF
--- a/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvscGenerationConfig.java
+++ b/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvscGenerationConfig.java
@@ -90,7 +90,7 @@ public class AvscGenerationConfig {
      */
     public static final AvscGenerationConfig CANONICAL_ONELINE = new AvscGenerationConfig(
         false, false, false, Optional.of(Boolean.FALSE), false, false,
-    false, false, false, false, true, false);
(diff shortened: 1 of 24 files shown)


dg-builder (changes requested):
Can we instead split this into 2 classes? `LegacyAvscWriter` and `AvscWriter`? `AvscWriter.java` is getting very long (and a lot of this code is copy-pasta

li-ukumar:
> Can we instead split this into 2 classes? `LegacyAvscWriter` and `AvscWriter`? `AvscWriter.java` is getting very long (and a lot of this code is copy-pasta

I suggest keeping AvscWriter intact and controlling what we write using configs and plugins. 
Also splitting avscwriter into parts would create quite a bit of duplicate as well

dg-builder (approved):

xiaoyu-yang-gh (approved):

PR #492, opened 2023-05-12
Support resolver path in SchemaBuilder

xiaoyu-yang-gh:
Add resolverPath to **SchemaBuilder** to allow user specify dependent Avro schemas.  User can add avsc files, directories that contain avsc files, Jar or Zip files that contain avsc files. Non-avsc files are ignored.

**ResolverPathSchemaSet** is implemented to load avro schemas from resolverPath, store the schemas in a map, and provide the method to lookup a schema by the fullname.

Test:
(1) Add unit test for ResolverPathSchemaSet for the following cases:
     - avsc file in resolverPath;
     - directory in resolverPath, directory contains non-avsc files;
     - Jar file in resolverPath, Jar contains non-avsc files;
     - Zip file in resolverPath; Zip contains non-avsc files;

(2) Add unit test method for SchemaBuilder, test the schema with dependencies can be build with the resolver path, java files can be generated.

(3) test by executing SchemaBuilder from user side.


avro-builder/builder/src/main/java/com/linkedin/avroutil1/builder/operations/codegen/own/AvroUtilCodeGenOp.java
```diff
@@ -100,10 +102,13 @@ public void run(OperationContext opContext) throws Exception {
...
 
-    //build a classpath SchemaSet if classpath (cp) lookup is turned on
-    ClasspathSchemaSet cpLookup = null;
-    if (config.isIncludeClasspath()) {
-      cpLookup = new ClasspathSchemaSet();
+    SchemaSet lookupSchemaSet = null;
+    if (config.getResolverPath() != null) {
+      //build a resolver path SchemaSet if ResolverPath is set
+      lookupSchemaSet = new ResolverPathSchemaSet(config.getResolverPath());
+    } else if (config.isIncludeClasspath()) {
```
radai-rosenblatt, line 109: we either document that CP and path are mutually exclusive, or we support both?

radai-rosenblatt: and if we choose to make them mutually exclusive - providing both arguments to the main() method should fail

radai-rosenblatt (approved):

PR #494, opened 2023-06-03
fix fixed literals

dg-builder:
AvroFixedLiteral equality was broken (before we were doing reference-level equality, not value-level equality).

UT was failing prior to the change in `AvroFixedLiteral`.


radai-rosenblatt (approved):
+1, good find

PR #495, opened 2023-06-05
handle recursive collections in RRG

dg-builder:
Recursive collections were throwing `StackOverflowError`s . This PR fixes that.

### DIFF
--- a/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/RandomRecordGenerator.java
+++ b/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/RandomRecordGenerator.java
@@ -113,6 +113,9 @@ public class RandomRecordGenerator {
         }
         return record;
       case ARRAY:
+        if (context.hasHitMaxDepth()) {
+          return new GenericData.Array<>(0, of);
+        }
         context.pushPath(of);
         size = random.nextInt(11); //[0, 10]
         GenericData.Array<Object> array = new GenericData.Array<>(size, of);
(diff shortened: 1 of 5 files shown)


helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/RecordGenerationContext.java
```diff
@@ -17,11 +17,14 @@
 public class RecordGenerationContext {
   private final RecordGenerationConfig config;
 
+  private final int MAX_DEPTH = 100;
+  private boolean hasHitMaxDepth = false;
```
radai-rosenblatt, line 21: whats the value of memoizing this now, given that the computation is fairly trivial?

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/RandomRecordGenerator.java
```diff
@@ -140,6 +146,11 @@ private Object newRandomGeneric(Schema of, RecordGenerationContext context) {
         }
         return map;
       case UNION:
+        if (context.hasHitMaxDepth() && of.getTypes()
```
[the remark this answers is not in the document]
dg-builder: ah you're right : `AvroSchemaUtil.isNullAValidDefaultForSchema(of)`

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/RecordGenerationContext.java
dg-builder: We may need to make this a bit more complex though...we should only stop generation for recursive schemas, not all schemas

dg-builder: added check directly on recursive schema depth (and memoized it)

radai-rosenblatt (approved):
works well enough. can be extended later

PR #499, opened 2023-07-10
Optimize enum field deserialization

### LINKED ISSUE #302: cache Enum.values() in fast-avro deserializers
when deserializing fields of type enum, for specific generated enum classes, the JVM does not cache Enum.vaules() and this can be a significant cause of memory churn and GC.

ClassValue could be used to cache the values() of enum classes

kcepmada:
BUG=FI-13182
Closes #302

ClassValue lazily caches result of Enum#values for the JVM to avoid unnecessary memory overhead.

odp download --event-id 1094224
* com.linkedin.followfeed.partitioner.graph.ContentQualityClassification[].values()
  Total Allocation: 20.5G (3.58%)


avro-codegen/src/main/java/com/linkedin/avroutil1/codegen/SpecificRecordGeneratorUtil.java
```diff
@@ -93,8 +93,25 @@ public class SpecificRecordGeneratorUtil {
       CompatibleSpecificRecordBuilderBase.class.getName()
   ));
 
+  private static final ClassValue<Enum<?>[]> ENUM_CONSTANTS = new ClassValue<Enum<?>[]>() {
```
radai-rosenblatt, line 96: putthing this here means generated classes will have a runtime dependency on the avro-codegen module. we'd like to keep helper as the only RUNTIME code dependency, so you should move this utility method (and the ClassValue backing it) to a util class in the helper module

avro-codegen/src/test/java/com/linkedin/avroutil1/codegen/SpecificRecordClassGeneratorTest.java
```diff
@@ -22,176 +21,43 @@
 
 
 public class SpecificRecordClassGeneratorTest {
-
-  @Test
```
radai-rosenblatt, line 26: what happened to all these tests?

avro-codegen/src/main/java/com/linkedin/avroutil1/codegen/SpecificRecordGeneratorUtil.java
kcepmada: Moved to com.linkedin.avroutil1.Enums inside the helper module.

avro-codegen/src/test/java/com/linkedin/avroutil1/codegen/SpecificRecordClassGeneratorTest.java
kcepmada: All still exist just they now share a parameterized implementation after collapsing redundant driver code. Originated from trying to understand and enhance coverage before moving on due to existing generated object file compilation issues. Left as a craftsmanship/readability improvement but happy to drop if preferred.

radai-rosenblatt (approved):

avro-codegen/src/test/java/com/linkedin/avroutil1/codegen/SpecificRecordClassGeneratorTest.java
kcepmada: Point taken. My bad, was lazy after fighting recent environment changes.

PR #501, opened 2023-07-31
throw better error msg on default parse failure

dg-builder:
When writing an `AvroSchema`, throw an exception with a clearer error message

### DIFF
--- a/parser/src/main/java/com/linkedin/avroutil1/parser/avsc/AvscParser.java
+++ b/parser/src/main/java/com/linkedin/avroutil1/parser/avsc/AvscParser.java
@@ -400,10 +400,11 @@ public class AvscParser {
                             }
                             LiteralOrIssue defaultValueOrIssue = parseLiteral(fieldDefaultValueNode, defaultValueExpectedSchema, fieldName.getValue(), context);
                             if (defaultValueOrIssue.getIssue() == null) {
-                                //TODO - allow parsing default values that are branch != 0 (and add an issue)
                                 defaultValue = defaultValueOrIssue.getLiteral();
                             } else {
                                 context.addIssue(defaultValueOrIssue.getIssue());
+                                defaultValue = new AvscUnparsedLiteral(fieldDefaultValueNode,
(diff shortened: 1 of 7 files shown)


parser/src/main/java/com/linkedin/avroutil1/parser/avsc/AvscParser.java
```diff
@@ -400,10 +400,11 @@ private AvroNamedSchema parseNamedSchema(
                             }
                             LiteralOrIssue defaultValueOrIssue = parseLiteral(fieldDefaultValueNode, defaultValueExpectedSchema, fieldName.getValue(), context);
                             if (defaultValueOrIssue.getIssue() == null) {
-                                //TODO - allow parsing default values that are branch != 0 (and add an issue)
```
dg-builder, line 403: default values for branches != 0 are not allowed per the avro spec

parser/src/main/java/com/linkedin/avroutil1/parser/avsc/AvscParser.java
```diff
@@ -400,10 +400,11 @@ private AvroNamedSchema parseNamedSchema(
...
                                 defaultValue = defaultValueOrIssue.getLiteral();
                             } else {
                                 context.addIssue(defaultValueOrIssue.getIssue());
+                                defaultValue = new AvscUnparsedLiteral(fieldDefaultValueNode,
+                                    defaultValueOrIssue.getIssue().getLocation());
```
dg-builder, line 407: We don't really handle issues properly so better to set the `defaultValue` to an unparsed literal to make sure the schema throws when trying to write

parser/src/main/java/com/linkedin/avroutil1/parser/avsc/AvscParser.java
radai-rosenblatt: true, but could have parsed the default value itself for any errors, in theory.

radai-rosenblatt (approved):

PR #504, opened 2023-08-15
Fix class generation for nested map field

nisargthakkar:
Records for schemas with nested maps fail to be when deserialized with fastserde. This PR adds a test case (`FastNestedMapTest.java`) to validate behavior


fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGeneratorBase.java
```diff
@@ -97,7 +97,7 @@ protected ListIterator<Symbol> actionIterator(FieldAction action) {
     while (actionIterator.hasNext()) {
       Symbol symbol = actionIterator.next();
 
-      if(Symbol.Kind.REPEATER.equals(symbol.kind) && "array-end".equals(((Symbol.Repeater)symbol).end.toString())) {
+      if(Symbol.Kind.REPEATER.equals(symbol.kind)) {
```
nisargthakkar, line 100: This change is not really needed, but I feel it should include `map-end` too. Let me know if you'd like me to remove this

Issue #257, opened 2021-11-24
customDecode() method on generated classes with 1000 properties explodes


radai-rosenblatt:
generates too much code, JVM cant handle more than 64K bytecode per method.

need to split it into smaller methods

PR #507, opened 2023-08-16
Codegen : Custom decode method broken into subMethods

li-ukumar:
## What
This change will break up generated customDecode method into small subMethods while keeping the functionality consistent.

## Why
Since generated customDecode method can be large for larger schemas, it could exceed size limit 64k. Splitting method into smaller subMethods avoids it.

Issue: https://github.com/linkedin/avro-util/issues/257

## Test
Added UT with a 1000 field record round trip ser/deserialization.
`Errors out on current master`
Screenshot 2023-08-16 at 5 00 10 PM

### DIFF
--- a/avro-codegen/src/main/java/com/linkedin/avroutil1/codegen/SpecificRecordClassGenerator.java
+++ b/avro-codegen/src/main/java/com/linkedin/avroutil1/codegen/SpecificRecordClassGenerator.java
@@ -961,7 +961,7 @@ public class SpecificRecordClassGenerator {
 
   private void addCustomDecodeMethod(MethodSpec.Builder customDecodeBuilder, AvroRecordSchema recordSchema,
       SpecificRecordGenerationConfig config, TypeSpec.Builder classBuilder) {
-    int blockSize = 5, fieldCounter = 0, chunkCounter = 0;
+    int blockSize = 25, fieldCounter = 0, chunkCounter = 0;
     // reset var counter
     sizeValCounter = -1;
     customDecodeBuilder.addStatement(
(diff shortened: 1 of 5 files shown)


avro-codegen/src/main/java/com/linkedin/avroutil1/codegen/SpecificRecordClassGenerator.java
```diff
@@ -959,41 +960,78 @@ private String getMethodNameForFieldWithPrefix(String prefix, String fieldName)
...
             "org.apache.avro.Schema.Field[] fieldOrder = (com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper.areFieldsReordered(getSchema())) ? in.readFieldOrder() : null")
         .beginControlFlow("if (fieldOrder == null)");
-    for(AvroSchemaField field : recordSchema.getFields()) {
-      String escapedFieldName = getFieldNameWithSuffix(field);
-      customDecodeBuilder.addStatement(getSerializedCustomDecodeBlock(config, field.getSchemaOrRef().getSchema(),
-          field.getSchemaOrRef().getSchema().type(), "this." + replaceSingleDollarSignWithDouble(escapedFieldName),
-          "this." + replaceSingleDollarSignWithDouble(escapedFieldName), StringUtils.EMPTY_STRING));
+
+    while (fieldCounter < recordSchema.getFields().size()) {
+      String chunkMethodName = "customDecodeIfChunk" + chunkCounter;
```
radai-rosenblatt, line 972: nit - why call the method "customDecodeIfChunk0()" ?
as oposed to customDecodeChunk0() ? what does the "if" stand for?

li-ukumar: Screenshot 2023-08-16 at 4 35 08 PM

The generated method has `if` and `else` chunks.

PR #508, opened 2023-08-17
Tweaked the fast-avro schema fingerprinting logic

FelixGV:
It used to leverage the Avro fingerprinting, which is based on the "parsing canonical form" of the schema, but this ignores properties such as which class to use for String deserialization. This is a problem in cases where we wish to deserialize the same schema both with and without Java Strings, since the cache will return the same deserializer for both, ultimately leading to class cast issues.

The new approach is to simply take the hash code of the full string of the schema.

### DIFF
--- a/fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/Utils.java
+++ b/fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/Utils.java
@@ -2,7 +2,6 @@ package com.linkedin.avro.fastserde;
 
 import com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper;
 import com.linkedin.avroutil1.compatibility.AvroVersion;
-import com.linkedin.avroutil1.compatibility.SchemaNormalization;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileReader;
@@ -39,7 +38,7 @@ public class Utils {
   }
 
   // Cache the mapping between Schema and the corresponding fingerprint
-  private static final Map<Schema, Long> SCHEMA_IDS_CACHE = new ConcurrentHashMap<>();
+  private static final Map<Schema, Integer> SCHEMA_IDS_CACHE = new ConcurrentHashMap<>();
 
   private Utils() {
   }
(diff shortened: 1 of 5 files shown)


fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/Utils.java
```diff
@@ -108,13 +107,14 @@ public static String fixSeparatorsToMatchOS(String path) {
...
    * @param schema a schema
    * @return fingerprint for the given schema
    */
-  public static Long getSchemaFingerprint(Schema schema) {
-    Long schemaId = SCHEMA_IDS_CACHE.get(schema);
+  public static int getSchemaFingerprint(Schema schema) {
+    Integer schemaId = SCHEMA_IDS_CACHE.get(schema);
     if (schemaId == null) {
-      schemaId = SchemaNormalization.parsingFingerprint64(schema);
+      schemaId = schema.toString().hashCode();
```
dg-builder, line 117: Two reasons to avoid this approach:
1. `Schema#toString()` is not stable. Long story, but it even has bugs under old versions of avro (see [AVRO-702](https://issues.apache.org/jira/browse/AVRO-702))
2. `String#hashCode()` isn't necessarily stable either.

We should probably normalize the schema first (you can use  `AvroUtilSchemaNormalization` and write a `AvscWriterPlugin` (see the `FieldLevelPlugin` example - something close to that should work for your use case) to include the junk json you want. [Here's a test](https://github.com/linkedin/avro-util/blob/07050eea0cccc9f665e80a870d6ebc5e130533ed/helper/tests/helper-tests-19/src/test/java/com/linkedin/avroutil1/compatibility/avro19/AvroUtilSchemaNormalizationTest19.java#L342-L361) case you can follow as an example).

However, you can also just use [AvscWriter](https://github.com/linkedin/avro-util/blob/07050eea0cccc9f665e80a870d6ebc5e130533ed/helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/AvscWriter.java) to serialize the schema if you don't need to normalize.

FelixGV: What do you mean by stable? If you mean that it will always yield the same value regardless of Avro version or JVM vendor/version, then that doesn't really matter. We only need this id to be stable within the scope of one runtime of the JVM.

That being said, I can still make it stable in case this ends up being used beyond one runtime in the future.

FelixGV: FYI @dg-builder, I have made the ID generation stable in my latest commit.

dg-builder (approved):

FelixGV:
Thanks for reviewing @dg-builder !

PR #509, opened 2023-08-24
Fast-Avro now supports logical types

krisso-rtb:


krisso-rtb:
Will be re-created later after internal code review (and cleanup).

PR #511, opened 2023-08-29
[fastserde] Fix issue with generating schema fingerprint when NaN is default value.

krisso-rtb:


helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/Jackson1Utils.java
```diff
@@ -210,20 +210,19 @@ public static boolean isRoundNumber(JsonNode node) {
...
     JsonNode defaultValue = field.defaultValue();
-    BigDecimal numericValue = defaultValue.getDecimalValue();
     Schema schema = field.schema();
     // a default value for a union, must match the first element of the union
     Schema.Type defaultType = schema.getType() == UNION ? schema.getTypes().get(0).getType() : schema.getType();
 
     switch (defaultType) {
       case INT:
-        if (!isAMathematicalInteger(numericValue)) {
+        if (!isAMathematicalInteger(defaultValue.getDecimalValue())) {
```
FelixGV, line 219: Is this the fix? I don't understand... isn't the code equivalent?

krisso-rtb: If `defaultType` is `DOUBLE` or `FLOAT` and `genericDefaultValue` represents `NaN` then calling `genericDefaultValue.decimalValue()` is
- not needed
- throws `NumberFormatException`

In other words it's called too early.

FelixGV: Oooh I see. Ok, cool. Thanks for explaining, good catch.

FelixGV (approved):
Thanks!

PR #514, opened 2023-09-08
[fast-avro] Added logic to split large method for fast serializer

gaojieliu:
In the past, we updated fast deserializer to support large method splitting, and this code change is leveraging the similar logic to split large method in fast serializer.


majisourav99 (approved):
LGTM, can you update the title of PR method->record.

gaojieliu:
> LGTM, can you update the title of PR method->record.

It is large method actually.
Large method can't be optimized by JIT by default.

PR #516, opened 2023-09-14
[fastserde] Logical types are now supported for avro version 1.9+

krisso-rtb:
The goal of this PR is to allow `fastserde` (de-)serializing records with logical-types.

I realize it's a quite big change so we did our best to verify it works fine... and no new bugs are added :)

----

Commits guide:
1. `Adding unit tests to verify LogicalTypes support.` - TDD approach, and failing tests are marked with `@Ignored`
2. `minor code cleanup` - self explanatory
3. `Generated classes - BEFORE` - temporarily adding all generated classes before the main changes
4. `modelData (GenericData or SpecificData instance) passed to FastSerdeBase` - should be part of the main commit (5.) however it's extracted as separate commit to make the main commit cleaner
5. `[fast-avro] Logical types are now supported by fast-avro in version 1.9+` - **the heart of this PR**, tests from (1.) are unignored.
6. `Generated fast-avro classes AFTER (logical-types related).` - temporary commit to see how main commit affected generated classes
7. `Generated fast-avro classes AFTER (side effects of supporting logical-types).` - same as above, could be considered as "side effects"
8. `Generated classes moved back to .gitignore` - drops changes from all temporary commits (3, 6 and 7)

----

Due to limitations of current Gradle's config it's impossible to generate Avro classes with **custom** logicalTypes. To verify this PR works fine in that case as well I created Unit Test in external (dedicated) project:
- https://github.com/krisso-rtb/avro-custom-logical-types/blob/master/custom-logical-types-used/src/test/java/com/rtbhouse/custom/logical/types/CustomLogicalTypesTest.java

which uses dependency built locally from this PR's branch:
```
<dependency>
    <groupId>com.linkedin.avroutil1</groupId>
    <artifactId>avro-fastserde</artifactId>
    <version>0.2.3-SNAPSHOT</version>
</dependency>
```


fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -184,147 +186,199 @@ public static boolean isFastDeserializer(FastDeserializer deserializer) {
...
    * @return specific-class aware avro {@link FastDeserializer}
    */
-  public FastDeserializer<?> getFastSpecificDeserializer(Schema writerSchema, Schema readerSchema) {
+  public FastDeserializer<?> getFastSpecificDeserializer(Schema writerSchema, Schema readerSchema, SpecificData modelData) {
     String schemaKey = getSchemaKey(writerSchema, readerSchema);
     FastDeserializer<?> deserializer = fastSpecificRecordDeserializersCache.get(schemaKey);
 
     if (deserializer == null) {
-      AtomicBoolean status = new AtomicBoolean(false);
+      AtomicBoolean fastDeserializerMissingInCache = new AtomicBoolean(false);
```
FelixGV, line 212: Thanks... that's much more clear.

fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -752,8 +762,9 @@ private void processArray(JVar arraySchemaVar, final String name, final Schema a
       action = FieldAction.fromValues(arraySchema.getElementType().getType(), false, EMPTY_SYMBOL);
     }
 
-    final JVar arrayVar = action.getShouldRead() ? declareValueVar(name, effectiveArrayReaderSchema, parentBody, true, false, true) : null;
-    /**
+    final boolean useLogicalType = action.getShouldRead() && logicalTypeEnabled(effectiveArrayReaderSchema.getElementType());
```
FelixGV, line 765: Seems like `useLogicalType` is only ever used when `action.getShouldRead() == true`, so it might not matter to for `useLogicalType` to have `action.getShouldRead() &&`... but if you think that makes the code more defensive or something like that it could be fine to leave it as is I guess.

fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/SchemaAssistant.java
```diff
@@ -444,6 +464,61 @@ public JClass findStringClass(Schema schema) {
     return outputClass;
   }
 
+  boolean logicalTypeEnabled(Schema schema) {
+//    return modelData != null && schema != null && Utils.isLogicalTypeSupported() && schema.getLogicalType() != null;
+    // TODO uuid in 1.10 no supported ??
```
FelixGV, line 469: Hmm, weird... I see UUIDs mentioned in the spec as far back as 1.9... but maybe in the actual implementation it's not properly supported until 1.11? IDK... this is the kind of minute detail @radai-rosenblatt would know :D ...

fastserde/avro-fastserde-tests110/build/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_10/FastSerdeLogicalTypesDefined_GenericDeserializer_229156053_229156053.java
```diff
@@ -14,13 +19,29 @@ public class FastSerdeLogicalTypesDefined_GenericDeserializer_229156053_22915605
...
     private final Schema arrayOfUnionOfDateAndTimestampMillis0;
     private final Schema arrayOfUnionOfDateAndTimestampMillisArrayElemSchema0;
+    private final Schema arrayOfUnionOfDateAndTimestampMillisElemOptionSchema0;
+    private final Schema arrayOfUnionOfDateAndTimestampMillisElemOptionSchema1;
+    private final Schema logicalTypeSchema_1074306973 = Schema.parse("{\"type\":\"long\",\"logicalType\":\"timestamp-millis\"}");
 
-    public FastSerdeLogicalTypesDefined_GenericDeserializer_229156053_229156053(Schema readerSchema) {
+    public FastSerdeLogicalTypesDefined_GenericDeserializer_229156053_229156053(Schema readerSchema, GenericData modelData) {
         this.readerSchema = readerSchema;
+        this.modelData = modelData;
```
FelixGV: Do we need to pass in the `modelData` and hang on to it as an instance property? I don't see any use of it in the generated code... I could have missed it maybe?

FelixGV (commented):
Thanks for the big code drop! The commit breakdown definitely helps navigate the change more easily. It LGTM overall. I left some minor comments and questions.

fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/SchemaAssistant.java
krisso-rtb: Ah yes, you reminded me something.
I did an experiment to see how generated Avro classes change when avro version is downgraded:
- https://github.com/krisso-rtb/avro-logical-types-generated-classes/commits/master
This particular case with `UUID` is caught by commit `Avro version downgraded from 1.11.1 to 1.11.0`
- https://github.com/krisso-rtb/avro-logical-types-generated-classes/commit/81e0be0e222989f81a0fe7461f173084f927aea2
Basically we have:
```
private java.util.UUID uuidField;   // avro 1.11.1
private java.lang.String uuidField; // avro 1.11.0
```
from the same `.avsc` file.
(just take a look at diff for `FastSerdeLogicalTypesTest1.java`)

fastserde/avro-fastserde-tests110/build/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_10/FastSerdeLogicalTypesDefined_GenericDeserializer_229156053_229156053.java
krisso-rtb: Thanks! You seems to be right. I will revisit the code and remove if it unless it's somewhere needed.
It's probably leftover from an early implementation.

fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
krisso-rtb: You're welcome :)

fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
krisso-rtb: Additional check for `action.getShouldRead()` is required because in some cases `effectiveArrayReaderSchema` is `null`, e.g. when reader schema is different than writer schema.
`effectiveArrayReaderSchema` is not `null` when `action.getShouldRead()` is `true`.

fastserde/avro-fastserde-tests-common/src/test/java/com/linkedin/avro/fastserde/logical/types/LogicalTypesTestBase.java
```diff
@@ -0,0 +1,214 @@
...
+                .orElse(Collections.emptyList())
+                .forEach(toModelData::addLogicalTypeConversion);
+
+        fixConversionsIfAvro19(toModelData);
+
+        return toModelData;
+    }
+
+    protected <T extends GenericData> void fixConversionsIfAvro19(T modelData) {
+        if (AvroCompatibilityHelperCommon.getRuntimeAvroVersion() == AvroVersion.AVRO_1_9) {
```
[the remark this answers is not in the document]
krisso-rtb: OK, I revisited this method and `new TimeConversions.TimeMillisConversion()` is not needed (will be removed in fix-commit).

Other 3 conversions
```
new Conversions.DecimalConversion()
new TimeConversions.TimeMicrosConversion()
new TimeConversions.TimestampMicrosConversion()
```
must be manually added because they're missing. They're actually "lost" when downgrading avro from 1.10.2 to 1.10.1:
- https://github.com/krisso-rtb/avro-logical-types-generated-classes/commit/e271f7d9a1ea60fca325a3c0b9260e5068e168bd
but in avro-util/fastserde we observe the difference between `1.10.2` and `1.9.2`.

fastserde/avro-fastserde-tests110/build/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_10/FastSerdeLogicalTypesDefined_GenericDeserializer_229156053_229156053.java
krisso-rtb: You were right! Thanks again!
I pushed the fixes in 9th commit.
To avoid force-pushing whole branch (GH may end up with dangling comments) I re-created the commits with fixes:
- https://github.com/RTBHOUSE/avro-util/pull/8 (temp PR, just to see the results in generated classes)
Files in "AFTER" commits now have no modelData field.

fastserde/avro-fastserde-tests-common/src/test/java/com/linkedin/avro/fastserde/logical/types/InMemoryEncoder.java
```diff
@@ -13,6 +13,17 @@
...
+/**
+ * Helper class which wraps {@link BinaryEncoder} and its corresponding {@code OutputStream}
+ * ({@link ByteArrayOutputStream}) so that we can simplify some repeatable code just to:
+ * <pre>
+ * InMemoryEncoder encoder = new InMemoryEncoder(data.getSchema());
+ * fastSerializer.serialize(data, encoder);
+ * return encoder.toByteArray();
+ * </pre>
+ * Additionally, it prepends 10-bytes header to the output byte-array so that it can be easily compared with
+ * built-in serialization method: {@code someSpecificRecord.toByteBuffer().toByteArray()}.
```
FelixGV, line 25: So... is the point of the header that two records of different schemas that happen to encode to the same bytes will not be mistakenly treated as equal?

FelixGV (commented):
Thanks for simplifying the code in the latest commit.

I have a high-level question about the strategy for this work...

Given that which supported logical type varies by Avro version (even by patch versions, unfortunately...), I wonder what is the ideal way to approach this?

I believe the approach you took in this PR is to take all of the logical types supported in the latest version of Avro, and if the runtime is Avro 1.9 or later, then use all of these logical types (I assume this is the case because of the `missingConversions` that you are registering into the `modelData`...). First of all, is this the right understanding of the current implementation? If it is, then I have some concerns, namely:

1. If using a SpecificRecord class that was generated in a version of Avro older than the latest, which is missing some of the logical types, then would we not have a class cast exception when trying to put an object of the latest logical type into a field that was generated to not have such type?
2. If using a GenericRecord, in theory we can put anything in any fields since it's just an array of Objects internally... however, during the period of time when fast-avro is "warming up" (i.e. before the serde compilation finishes), then we'll be falling back on vanilla avro... so does that mean the user will see older types for a brief period of time and later on see the latest logical types?

Both of the above behaviors seem wrong... so I wonder if we should instead do some runtime introspection to figure out what the right type to use is, at the time of the serde code gen, and then use the right type accordingly in the generated code. For a `SpecificRecord` serde, the right type could be figured out by using reflection to introspect the fields of the `SpecificRecord` class. For `GenericRecord`, perhaps we could synthetically generate a payload based on the schema, deserialize it with vanilla Avro, and introspect what type each field is...

It is troublesome, but I think it would avoid the issues mentioned above.

Alternatively, we could say that fast-avro only supports logical types on the most recent of all Avro versions, and for that version it supports all the types... This would be roughly equivalent to the current PR, but we would just bump the minimum Avro version to activate the logical type behavior... Easier to implement, though technically still incorrect in terms of Avro 1.9 and 1.10, but at least it's easier to explain to users what to expect.

Maybe I'm twisting myself into knots over this and it need not be this complicated... IDK. What do you think?

krisso-rtb:
> Thanks for simplifying the code in the latest commit.
> 
> I have a high-level question about the strategy for this work...
> 
> Given that which supported logical type varies by Avro version (even by patch versions, unfortunately...), I wonder what is the ideal way to approach this?
> 
> I believe the approach you took in this PR is to take all of the logical types supported in the latest version of Avro, and if the runtime is Avro 1.9 or later, then use all of these logical types (I assume this is the case because of the `missingConversions` that you are registering into the `modelData`...). First of all, is this the right understanding of the current implementation? If it is, then I have some concerns, namely:
> 
> 1. If using a SpecificRecord class that was generated in a version of Avro older than the latest, which is missing some of the logical types, then would we not have a class cast exception when trying to put an object of the latest logical type into a field that was generated to not have such type?
> 2. If using a GenericRecord, in theory we can put anything in any fields since it's just an array of Objects internally... however, during the period of time when fast-avro is "warming up" (i.e. before the serde compilation finishes), then we'll be falling back on vanilla avro... so does that mean the user will see older types for a brief period of time and later on see the latest logical types?
> 
> Both of the above behaviors seem wrong... so I wonder if we should instead do some runtime introspection to figure out what the right type to use is, at the time of the serde code gen, and then use the right type accordingly in the generated code. For a `SpecificRecord` serde, the right type could be figured out by using reflection to introspect the fields of the `SpecificRecord` class. For `GenericRecord`, perhaps we could synthetically generate a payload based on the schema, deserialize it with vanilla Avro, and introspect what type each field is...
> 
> It is troublesome, but I think it would avoid the issues mentioned above.
> 
> Alternatively, we could say that fast-avro only supports logical types on the most recent of all Avro versions, and for that version it supports all the types... This would be roughly equivalent to the current PR, but we would just bump the minimum Avro version to activate the logical type behavior... Easier to implement, though technically still incorrect in terms of Avro 1.9 and 1.10, but at least it's easier to explain to users what to expect.
> 
> Maybe I'm twisting myself into knots over this and it need not be this complicated... IDK. What do you think?

1. I assumed **it's the responsibility of fast-avro user to provide correct `modelData`** (which is either `GenericData` or `SpecificData`). If specific record class has been generated years ago then we must provide `specificData` with no conversions (or simply `null`) to `FastSpecificDatumReader`.
When it comes to (de)serialization, conversion classes can't be deducted from the schema. That's why we need `modelData`. What's more - implementation of custom conversion class may change between these two points in time:
- generation of Avro specific record class
- actual (de)serialization
**Providing correct `modelData` is as important as providing correct `schema`.**

2. During "warm up" period we use standard Avro (de)serializers which also receive `modelData`, e.g.
```
protected FastDeserializer<T> getRegularAvroImpl(Schema writerSchema, Schema readerSchema, GenericData modelData) {
  return new FastSerdeCache.FastDeserializerWithAvroGenericImpl<>(writerSchema, readerSchema, modelData);
}
```
so they use the same conversion classes. It's implicitly tested but just in case I will explicitly add these classes to `LogicalTypesTestBase` as well:
- `ColdGenericDatumReader`
- `ColdSpecificDatumReader`
- `FastDeserializerWithAvroGenericImpl`
- `FastDeserializerWithAvroSpecificImpl`
- `FastSerializerWithAvroGenericImpl`
- `FastSerializerWithAvroSpecificImpl`

The hacks/tricks in unit tests, represented by `fixConversionsIfAvro19()` method, come from the fact generated Avro classes are inconsistent.
E.g. `FastSerdeLogicalTypesTest1` in Avro 1.9 generates field for `"logicalType": "time-micros"` but corresponding conversion is not added to `MODEL$`
```
static {
    MODEL$.addLogicalTypeConversion(new org.apache.avro.data.TimeConversions.DateConversion());
    MODEL$.addLogicalTypeConversion(new org.apache.avro.data.TimeConversions.TimestampMillisConversion());
    MODEL$.addLogicalTypeConversion(new org.apache.avro.data.TimeConversions.TimeMillisConversion());
  }
  ...
private java.time.LocalTime timeMicrosField;
```

FelixGV (approved):
Thanks for continuing to push on this. I discussed it offline with @radai-rosenblatt, and we have a minor concern, but I don't think it should be a blocker given that this new capability is opt-in.

The concern is that there is maybe too much trust that the users will do the right thing. In our internal ecosystem, we usually cannot make assumptions about which version of Avro a SR was generated from and need to be able to deal with all of them. Furthermore, some SR may be embedded in libraries and mixed with other libraries or main application code and each of these sources of code may have been touched by different versions of Avro. We think there may be ways to make the work begun here even more robust by introspecting types at serde code gen time, rather than relying on the user to pass in the right mappings in their `modelData`... but I don't think that should block this from being merged, since as I said, it is opt-in.

It is great to see support for logical types finally added 😄 🎉 ...

PR #517, opened 2023-09-19
[fast-avro] Added a few serde optimization

gaojieliu:
Serializer:
This code change will try to reuse the backed bytes if the float list is 'BufferBackedPrimitiveFloatList' when writing float list. If an instance of `BufferBackedPrimitiveFloatList` is changed after deserialization: `readPrimitiveFloatArray`, Fast Serializer won't use the backed bytes because of the divergence.

Deserializer:
Use `reset` instead of `clear` when reusing `GenericArray` since `reset` is cheaper than `clear` and the behavior difference is that `reset` won't nullify the elements in current array, but just resize the array length to be 0.

### DIFF
--- a/fastserde/avro-fastserde-tests-common/src/test/java/com/linkedin/avro/fastserde/FastGenericSerializerGeneratorTest.java
+++ b/fastserde/avro-fastserde-tests-common/src/test/java/com/linkedin/avro/fastserde/FastGenericSerializerGeneratorTest.java
@@ -12,6 +12,7 @@ import com.linkedin.avro.fastserde.generated.avro.JustSimpleEnum;
 import com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
+import java.io.IOException;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.nio.ByteBuffer;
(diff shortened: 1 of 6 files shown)


fastserde/avro-fastserde-tests-common/src/test/java/com/linkedin/avro/fastserde/FastGenericSerializerGeneratorTest.java
```diff
@@ -562,6 +563,81 @@ public double getPrimitive(int index) {
...
+    class TestBufferBackedPrimitiveFloatList extends BufferBackedPrimitiveFloatList {
+
+      boolean writeFloatsCalled = false;
+      boolean writeFloatsByBackedBytesCalled = false;
+      public TestBufferBackedPrimitiveFloatList(BufferBackedPrimitiveFloatList floatList) {
+        super(0);
+        floatList.copyInternalState(this);
+      }
+
+      public void reset() {
```
FelixGV, line 588: Can we call this `resetFlags` or something like that? I was a bit confused because I was thinking about the `GenericArray`'s `reset` API which is also involved in this commit, but unrelated to the code here...

fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/BufferBackedPrimitiveFloatList.java
```diff
@@ -61,6 +65,15 @@ public BufferBackedPrimitiveFloatList(Collection<Float> c) {
     byteBuffer = new CompositeByteBuffer(c != null);
   }
 
+  // For testing purpose
```
FelixGV, line 68: I suggest using JavaDoc rather than a double slash, and mentioning that this is a shallow copy.

fastserde/avro-fastserde/src/main/java11/com/linkedin/avro/fastserde/BufferBackedPrimitiveFloatList.java
```diff
@@ -60,6 +63,15 @@ public BufferBackedPrimitiveFloatList(Collection<Float> c) {
     }
   }
 
+  // For testing purpose
```
FelixGV, line 66: Same comment as above

fastserde/avro-fastserde/src/main/java11/com/linkedin/avro/fastserde/BufferBackedPrimitiveFloatList.java
```diff
@@ -250,11 +264,12 @@ public void add(int location, Float o) {
     System.arraycopy(elements, location, elements, location + 1, size - location);
     elements[location] = o;
     size++;
+    changed = true;
   }
 
   @Override
   public Float set(int i, Float o) {
-    return set(i, o);
+    return setPrimitive(i, o);
```
FelixGV, line 272: Does this fix a stackoverflow 😅 ?

FelixGV (commented):
Great stuff, thanks Gaojie! Left just a few minor comments...

alex-dubrouski:
Should we wait until we finish benchmarking deserialization path?

FelixGV:
> Should we wait until we finish benchmarking deserialization path?

Why wait? We can always submit another PR if we discover other opportunities, right?

fastserde/avro-fastserde/src/main/java11/com/linkedin/avro/fastserde/BufferBackedPrimitiveFloatList.java
gaojieliu: Yep!

FelixGV (approved):
LGTM! Thanks!

PR #519, opened 2023-10-09
Adding fail-fast feature to fast-serde

kbarczynski-rtbh:
This PR aims to add "fail-fast" feature to the fastserde.

Normally when we have datum writer:
```
DatumWriter<T> datumWriter = new FastSpecificDatumWriter<>(schema, specificData);
```
and it's used in a loop as:
```
datumWriter.write(record, binaryEncoder);
```
the `datumWriter` under the hood uses standard avro writer until fast-serializer is generated on the background-thread.

Currently if such generation fails due to any reason we have just one log entry and `datumWriter` still uses standard avro writer.
We found this approach is sometimes too gentle because it hides an issue for a long time. It's easy to miss one WARN log-entry.

That's why we think "fail-fast" approach would be a nice feature. It can be turned on by setting one of system properties
- `avro.fast.serde.failfast`
- `avro.fast.serde.failfast.supplier`

By default it's disabled so that it's backward-compatible.

----

Commits:
1. `Code cleanup`
- `private Optional<String> compileClassPath;` changed to `private String compileClassPath`;
- Resolving `compileClassPath` extracted to helper method (will be re-used for the new system property)
- other minor things
2. `bugfixes`
- `getFastGenericDeserializerAsync` no longer generated StackOverflow (shame on me!)
- Using `AvroCompatibilityHelper` in fallback-serializers to fix an issue with missing c-tor in Avro 1.4.
3. `Generation errors logged on ERROR level.`
- We can discuss here - I'm fine with dropping this commit if you think it should stay as WARN.
4. `[fastserde] Adding fail-fast feature.`
- The main reason why this PR was created.

----

Name "fail-fast" is a kind of industry standard however in this case it may cause confusion with "fast-serde" - we can use another name for this feature (fail-early / force-fastserde / etc.) or give it opposite meaning (e.g. ignore-generation-errors).

### DIFF
--- a/fastserde/avro-fastserde-tests-common/src/test/java/com/linkedin/avro/fastserde/FastSerdeCacheTest.java
+++ b/fastserde/avro-fastserde-tests-common/src/test/java/com/linkedin/avro/fastserde/FastSerdeCacheTest.java
@@ -1,11 +1,23 @@
 package com.linkedin.avro.fastserde;
 
+import com.linkedin.avro.fastserde.generated.avro.SimpleRecord;
 import com.linkedin.avro.fastserde.generated.avro.TestRecord;
+import com.linkedin.avroutil1.compatibility.AvroCompatibilityHelper;
+
+import java.io.ByteArrayOutputStream;
+import java.lang.reflect.Field;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.BiFunction;
+
 import org.apache.avro.Schema;
+import org.apache.avro.io.BinaryDecoder;
+import org.apache.avro.io.BinaryEncoder;
+import org.apache.avro.io.DatumReader;
+import org.apache.avro.io.DatumWriter;
+import org.apache.avro.specific.SpecificDatumWriter;
 import org.testng.Assert;
 import org.testng.annotations.Test;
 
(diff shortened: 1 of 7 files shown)


krisso-rtb:
Hi @FelixGV @gaojieliu 
Did you have a chance to look into this?
Just a kind reminder :)

fastserde/avro-fastserde-tests-common/src/test/java/com/linkedin/avro/fastserde/FastSerdeCacheTest.java
```diff
@@ -68,4 +80,81 @@ public void testBuildFastSpecificDeserializerWithCorrectClasspath() {
...
+  public void testSpecificDeserializationFailsFast() throws Exception {
+    deserializationShouldFailFast(FastSpecificDatumReader::new);
+  }
+
+  @Test(groups = "deserializationTest", timeOut = 5_000L,
+          expectedExceptions = UnsupportedOperationException.class,
+          expectedExceptionsMessageRegExp = "Fast generic deserializer could not be generated.")
+  public void testGenericDeserializationFailsFast() throws Exception {
+    deserializationShouldFailFast(FastGenericDatumReader::new);
+  }
```
gaojieliu, line 110: You can use data provider to avoid these test classes.
https://testng.org/doc/documentation-main.html#parameters-dataproviders

fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
```diff
@@ -89,44 +92,45 @@ public FastSerdeCache(Executor executorService, Supplier<String> compileClassPat
...
 
   /**
    *
    * @param executorService
    *            customized {@link Executor} used by serializer/deserializer compile threads
-   * @param compileClassPath
-   *            custom classpath as string
    */
+  public FastSerdeCache(Executor executorService) {
+    this(executorService, (String) null);
```
gaojieliu, line 104: Is this string type cast necessary?

gaojieliu (commented):
Looks good to me, thanks for the change!

fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastSerdeCache.java
krisso-rtb: Yes - we need to distinguish c-tors:
```
public FastSerdeCache(Executor executorService, Supplier<String> compileClassPathSupplier)
```
and 
```
public FastSerdeCache(Executor executorService, String compileClassPath)
```

fastserde/avro-fastserde-tests-common/src/test/java/com/linkedin/avro/fastserde/FastSerdeCacheTest.java
krisso-rtb: I was thinking about it but at the end I think current way is easier to read.
Please keep in mind we would need data providers (it's 2x2 not 4x1). These two methods have too much differences to unify them (without loosing readability):
```
serializationShouldFailFast
deserializationShouldFailFast
```

There other thing is data-providers would force explicit try-catch blocks instead of annotation's attributes:
```
expectedExceptions = UnsupportedOperationException.class,
expectedExceptionsMessageRegExp = "Fast specific deserializer could not be generated."
```

gaojieliu (approved):

krisso-rtb:
Thanks for merging PR to master. Would it be possible to have a new release? :)

krisso-rtb:
That's awesome!
Thank you!

PR #520, opened 2023-10-17
Added serde customization support in Fast-Avro

gaojieliu:
We have the following requirements:
For serialization, we would like to validate whether all the map fields are using the desired map type.
For deserialization, we would like to deserialize the map type into a special map impelementation for later use.

These customized requirements are not supported in the past because of the following reasons:
1. Fast classes generated are shared, so it is possible different users of the same schema may have different requirement.
2. For the same process, for different schema, the requirements can be different too.
3. No way to specify customized logic/data type when generating fast classes.

This PR adds a new functionality to specify customized logic and it is expandable and backward compatible.
DatumReaderCustomization : customization for read
DatumWriterCustomization : customization for write

Currently, these classes only support the requirements mentioned at the beginning.

How it works internally?
1. Each Fast DatumReader/DatumWriter constructor will take a new param for customization.
2. Each Fast DatumReader/DatumWriter will keep a local vanilla-Avro based implementation with customization support since the shared vanilla-Avro based implementation is still the default implementation.
3. Each generated Fast class will have a new param for customization in serialize/deserialize APIs.
4. Fast DatumReader/DatumWriter will call this new API with customization param of Fast classes.
5. The read/write API in Fast DatumReader/DatumWriter doesn't change, so it is backward compatible.

### DIFF
--- a/fastserde/avro-fastserde-tests-common/src/test/java/com/linkedin/avro/fastserde/FastDatumReaderTest.java
+++ b/fastserde/avro-fastserde-tests-common/src/test/java/com/linkedin/avro/fastserde/FastDatumReaderTest.java
@@ -1,8 +1,14 @@
 package com.linkedin.avro.fastserde;
 
+import com.linkedin.avro.fastserde.customized.DatumReaderCustomization;
 import com.linkedin.avro.fastserde.generated.avro.TestEnum;
 import com.linkedin.avro.fastserde.generated.avro.TestRecord;
 import java.io.IOException;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericData;
@@ -81,7 +87,7 @@ public class FastDatumReaderTest {
         (FastDeserializer<TestRecord>) cache.getFastSpecificDeserializer(TestRecord.SCHEMA$, faultySchema);
 
     Assert.assertNotNull(fastSpecificDeserializer);
-    Assert.assertEquals(fastSpecificDeserializer.getClass().getDeclaredMethods().length, 1);
+    Assert.assertFalse(fastSpecificDeserializer.isBackedByGeneratedClass());
   }
 
   @Test(groups = {"deserializationTest"})
@@ -133,5 +139,61 @@ public class FastDatumReaderTest {
         fastGenericDatumReader.read(null, FastSerdeTestsSupport.genericDataAsDecoder(record)).get("test"));
     Assert.assertTrue(fastGenericDatumReader.isFastDeserializerUsed(), "FastGenericDatumReader should be using"
         + " Fast Deserializer when the fast deserializer generation is done.");
+  }
+
+  @Test(groups = {"deserializationTest"})
+  public void testDatumReaderWithCustomization() throws IOException, ExecutionException, InterruptedException {
+    Schema recordSchema = createRecord("TestSchema",
+        createField("testInt", Schema.create(Schema.Type.INT)),
+        createMapFieldSchema("testMap", Schema.create(Schema.Type.STRING)));
+    /**
+     * Test with special map type: {@link java.util.concurrent.ConcurrentHashMap}.
+     */
+    DatumReaderCustomization customization = new DatumReaderCustomization.Builder()
+        .setNewMapOverrideFunc( (reuse, size) -> {
+          if (reuse instanceof ConcurrentHashMap) {
+            ((ConcurrentHashMap)reuse).clear();
+            return reuse;
+          } else {
+            return new ConcurrentHashMap<>(size);
+          }
+        })
+        .build();
+    // Check cold datum Reader
+    GenericRecord record = new GenericData.Record(recordSchema);
+    record.put("testInt", new Integer(100));
+    Map<Utf8, Utf8> testMap = new HashMap<>();
+    testMap.put(new Utf8("key1"), new Utf8("value1"));
+    testMap.put(new Utf8("key2"), new Utf8("value2"));
+    record.put("testMap", testMap);
+    FastGenericDatumReader<GenericRecord> fastGenericDatumReader = new FastGenericDatumReader<>(recordSchema, recordSchema, cache, null, customization);
+    GenericRecord deserializedRecordByColdDatumReader = fastGenericDatumReader.read(null, FastSerdeTestsSupport.genericDataAsDecoder(record));
+    Assert.assertEquals(deserializedRecordByColdDatumReader.get("testInt"), new Integer(100));
+    Assert.assertEquals(deserializedRecordByColdDatumReader.get("testMap"), testMap);
+    Assert.assertTrue(deserializedRecordByColdDatumReader.get("testMap") instanceof ConcurrentHashMap);
+
+    // Block the fast deserializer generation
+    fastGenericDatumReader.getFastDeserializer().get();
+    // Decode the record by fast datum reader
+    GenericRecord deserializedRecordByFastDatumReader = fastGenericDatumReader.read(null, FastSerdeTestsSupport.genericDataAsDecoder(record));
+    Assert.assertEquals(deserializedRecordByFastDatumReader.get("testInt"), new Integer(100));
+    Assert.assertEquals(deserializedRecordByFastDatumReader.get("testMap"), testMap);
+    Assert.assertTrue(deserializedRecordByFastDatumReader.get("testMap") instanceof ConcurrentHashMap);
+
+    // Test with an empty map
+    GenericRecord recordWithEmptyMap = new GenericData.Record(recordSchema);
+    recordWithEmptyMap.put("testInt", new Integer(200));
+    recordWithEmptyMap.put("testMap", Collections.emptyMap());
+    GenericRecord deserializedRecordWithEmptyMapByFastDatumReader = fastGenericDatumReader.read(null, FastSerdeTestsSupport.genericDataAsDecoder(recordWithEmptyMap));
+    Assert.assertEquals(deserializedRecordWithEmptyMapByFastDatumReader.get("testInt"), new Integer(200));
+    Assert.assertEquals(deserializedRecordWithEmptyMapByFastDatumReader.get("testMap"), Collections.emptyMap());
+    Assert.assertTrue(deserializedRecordWithEmptyMapByFastDatumReader.get("testMap") instanceof ConcurrentHashMap);
+
+    // Generate a new fast datum reader with the same schema, but without customization
+    FastGenericDatumReader<GenericRecord> fastGenericDatumReaderWithoutCustomization = new FastGenericDatumReader<>(recordSchema, cache);
+    GenericRecord deserializedRecordByFastDatumReaderWithoutCustomization = fastGenericDatumReaderWithoutCustomization.read(null, FastSerdeTestsSupport.genericDataAsDecoder(record));
+    Assert.assertEquals(deserializedRecordByFastDatumReaderWithoutCustomization.get("testInt"), new Integer(100));
+    Assert.assertEquals(deserializedRecordByFastDatumReaderWithoutCustomization.get("testMap"), testMap);
+    Assert.assertFalse(deserializedRecordByFastDatumReaderWithoutCustomization.get("testMap") instanceof ConcurrentHashMap);
   }
 }
--- a/fastserde/avro-fastserde-tests111/build/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_11/Array_of_TestRecord_SpecificDeserializer_1088113243_1088113243.java
+++ b/fastserde/avro-fastserde-tests111/build/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_11/Array_of_TestRecord_SpecificDeserializer_1088113243_1088113243.java
@@ -4,7 +4,6 @@ package com.linkedin.avro.fastserde.generated.deserialization.AVRO_1_11;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import com.linkedin.avro.api.PrimitiveBooleanList;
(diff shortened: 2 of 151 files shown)


FelixGV (commented):
Didn't finish reviewing, but LGTM so far. I'll finish later.

fastserde/avro-fastserde-tests-common/src/test/java/com/linkedin/avro/fastserde/FastDatumReaderTest.java
```diff
@@ -134,4 +140,60 @@ public void shouldCreateGenericDatumReader() throws IOException {
...
+    // Decode the record by fast datum reader
+    GenericRecord deserializedRecordByFastDatumReader = fastGenericDatumReader.read(null, FastSerdeTestsSupport.genericDataAsDecoder(record));
+    Assert.assertEquals(deserializedRecordByFastDatumReader.get("testInt"), new Integer(100));
+    Assert.assertEquals(deserializedRecordByFastDatumReader.get("testMap"), testMap);
+    Assert.assertTrue(deserializedRecordByFastDatumReader.get("testMap") instanceof ConcurrentHashMap);
+
+    // Test with an empty map
+    GenericRecord recordWithEmptyMap = new GenericData.Record(recordSchema);
+    recordWithEmptyMap.put("testInt", new Integer(200));
+    recordWithEmptyMap.put("testMap", Collections.emptyMap());
```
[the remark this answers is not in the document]
gaojieliu: yeah, we can consider doing this in OSS Venice.

FelixGV:
Thanks for adding back the generated classes @gaojieliu, but can we add them before the code change, so we can see the diff? i.e. these steps as separate commits:

1. Tweak the ignore file to add back the generated code (perhaps in just one Avro version, or all version... either way may be fine...).
2. Do the code change.
3. Changes to the generated code.
4. (Possibly optional) undo the tweak to the ignore file to re-ignore generated code...

On a separate note, @radai-rosenblatt, how strongly do you feel about not checking in the generated code? It makes it quite tedious, needing to do these git acrobatics in order to see the effect of the changes to the meta code...

gaojieliu:
> Thanks for adding back the generated classes @gaojieliu, but can we add them before the code change, so we can see the diff? i.e. these steps as separate commits:
> 
> 1. Tweak the ignore file to add back the generated code (perhaps in just one Avro version, or all version... either way may be fine...).
> 2. Do the code change.
> 3. Changes to the generated code.
> 4. (Possibly optional) undo the tweak to the ignore file to re-ignore generated code...
> 
> On a separate note, @radai-rosenblatt, how strongly do you feel about not checking in the generated code? It makes it quite tedious, needing to do these git acrobatics in order to see the effect of the changes to the meta code...

It will be great to checkin one version of generated classes, otherwise the PR effort will be too high..

fastserde/avro-fastserde-tests111/build/codegen/java/com/linkedin/avro/fastserde/generated/deserialization/AVRO_1_11/Array_of_TestRecord_SpecificDeserializer_1088113243_1088113243.java
```diff
@@ -527,39 +528,47 @@ private void populate_TestRecord11(com.linkedin.avro.fastserde.generated.avro.Te
...
         if (chunkLen6 > 0) {
-            Map<Utf8, com.linkedin.avro.fastserde.generated.avro.SubRecord> recordsMapReuse0 = null;
-            Object oldMap0 = TestRecord.get(25);
-            if (oldMap0 instanceof Map) {
-                recordsMapReuse0 = ((Map) oldMap0);
-            }
-            if (recordsMapReuse0 != (null)) {
-                recordsMapReuse0 .clear();
-                recordsMap0 = recordsMapReuse0;
+            if (((customization)!= null)&&((customization).getNewMapOverrideFunc()!= null)) {
```
FelixGV, line 544: Hmm... we know at code gen time whether these objects are null, right? It would be nice to eliminate this fork in the generated code, and generate only the path that will actually execute...

Or another way to do it might be that at code gen time, if these objects are null, then we populate them with a default impl which is essentially the same as the `else` branch below, then the generated code always calls the customization function.

LMK what you think...

FelixGV (commented):
Thanks for restructuring the PR's commits. I left one more comment.

fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/customized/DatumReaderCustomization.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+        if (retMap != null) {
+          retMap.clear();
+        } else {
+          /**
+           * Pure integer arithmetic equivalent of (int) Math.ceil(expectedSize / 0.75).
+           * The default load factor of HashMap is 0.75 and HashMap internally ensures size is always a power of two.
+           */
+          retMap = new HashMap<>((size * 4 + 2) / 3);
+        }
```
FelixGV, line 57: I don't think we need two separate ifs here... `null instanceof Map` is always false, so we can compress this code to:

```
if (old instanceof Map) {
  retMap = (Map)old;
  retMap.clear();
} else {
  retMap = new HashMap<>((size * 4 + 2) / 3);
}
```

fastserde/avro-fastserde/src/main/java/org/apache/avro/generic/CustomizedGenericDatumReader.java
```diff
@@ -0,0 +1,44 @@
...
+    this.customization = customization;
+  }
+
+
+  @Override
+  protected Object newMap(Object old, int size) {
+    if (customization.getNewMapOverrideFunc() != null) {
+      return customization.getNewMapOverrideFunc().apply(old, size);
+    }
+    return super.newMap(old, size);
```
FelixGV, line 37: We can make this the same as the generated code, right? Just always call:

```
return customization.getNewMapOverrideFunc().apply(old, size);
```

fastserde/avro-fastserde/src/main/java/org/apache/avro/generic/CustomizedGenericDatumWriter.java
```diff
@@ -0,0 +1,31 @@
...
+
+    this.customization = customization;
+  }
+
+  @Override
+  protected void writeMap(Schema schema, Object datum, Encoder out) throws IOException {
+    if (customization != null && customization.getCheckMapTypeFunction() != null) {
+      customization.getCheckMapTypeFunction().apply(datum);
+    }
+    super.writeMap(schema, datum, out);
```
FelixGV, line 29: Same here, we can always call it directly now, right? And the two specific classes below as well.

fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/customized/DatumWriterCustomization.java
```diff
@@ -0,0 +1,40 @@
...
+    this.checkMapTypeFunction = builder.checkMapTypeFunction;
+  }
+
+  public CheckMapTypeFunction getCheckMapTypeFunction() {
+    return checkMapTypeFunction;
+  }
+
+  public interface CheckMapTypeFunction {
+    void apply(Object o);
+  }
```
FelixGV, line 20: What is the goal of this function? To throw if the check fails?

Do we not have a use case where we want to transform the map (e.g. sort the keys) prior to serialization?

FelixGV (commented):
Thanks for iterating on this Gaojie. It LGTM overall. I have a few more minor comments.

Also, I think it would be good to make the float list customization part of this new framework (but it doesn't have to be part of this PR).

fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/customized/DatumWriterCustomization.java
gaojieliu: yeah, in OSS Venice, there is a safeguard when serializing the write compute result to make sure each map field will use an ordered map.
https://github.com/linkedin/venice/blob/main/clients/da-vinci-client/src/main/java/com/linkedin/davinci/serializer/avro/MapOrderPreservingGenericDatumWriter.java#L23

fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/customized/DatumReaderCustomization.java
gaojieliu: Sure.

fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/customized/DatumWriterCustomization.java
FelixGV: Ah ok, so we just fail there... I thought we were sorting on the fly... so I guess this is good enough to support our current use case. IDK if there would ever be a use case for on the fly transformation, so I guess we can leave it as is for now.

fastserde/avro-fastserde/src/main/java/org/apache/avro/generic/CustomizedGenericDatumReader.java
gaojieliu: Yes, we can now.
In the future, when we support more customization functions, we need to backfill all the missing functions to make sure each function won't be null, otherwise, this logic can break.

fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/customized/DatumWriterCustomization.java
gaojieliu: Well, ordered map doesn't mean it is sorted.
`IndexedHashMap` is not sorted, and it is more like a LinkedList to preserve the original order.

fastserde/avro-fastserde/src/main/java/org/apache/avro/generic/CustomizedGenericDatumReader.java
FelixGV: Yes, that is how I hope it will be... the setter on the customization object should reject null closures, and by default should be populated with the vanilla implementation.

fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/customized/DatumWriterCustomization.java
FelixGV: Actually, we do have a use case where we alter the written map, right here: https://github.com/linkedin/venice/blob/585036c9b0553c8a96653d9255c1e2dbd4bd93a3/internal/venice-client-common/src/main/java/org/apache/avro/generic/DeterministicMapOrderGenericDatumWriter.java#L46

FelixGV (approved):

fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/customized/DatumWriterCustomization.java
gaojieliu: Will address this in another PR.

krisso-rtb:
Hi @gaojieliu 
After upgrading to `0.3.21` we started observing errors:
```
java.lang.NullPointerException: Cannot invoke "org.apache.avro.Schema.equals(Object)" because "writer" is null
	at org.apache.avro.Schema.applyAliases(Schema.java:1890)
	at org.apache.avro.generic.GenericDatumReader.getResolver(GenericDatumReader.java:131)
	at org.apache.avro.generic.GenericDatumReader.read(GenericDatumReader.java:152)
	at com.linkedin.avro.fastserde.FastSerdeUtils$FastDeserializerWithAvroSpecificImpl.deserialize(FastSerdeUtils.java:68)
	at com.linkedin.avro.fastserde.FastGenericDatumReader.read(FastGenericDatumReader.java:126)
	at org.apache.avro.file.DataFileStream.next(DataFileStream.java:263)
	at org.apache.avro.file.DataFileStream.next(DataFileStream.java:248)
```

Presence of the new class (`FastSerdeUtils`) pointed us to this PR...
Has something changed?

We use
```
new FastSpecificDatumReader<>(null,  schema);
```
but it looks like `null` is not allowed anymore.

FelixGV:
Thanks for reporting... that is weird, since I don't see changes to the `FastSpecificDatumReader` constructors... but... what does it mean to have a null writer schema, though? Are you expecting the writer schema to default the reader schema in that case?

krisso-rtb:
OK, it's a bit more complex scenario :) 
```
Schema schema = SomeAvroClass.getClassSchema();
FastSpecificDatumReader<SomeAvroClass> datumReader = new FastSpecificDatumReader<>(null, schema);

Path path = Paths.get("/Users/kris/dev/data-file-with-schema-in-header.avro");
try (InputStream inputStream = Files.newInputStream(path);
    DataFileStream<SomeAvroClass> reader = new DataFileStream<>(inputStream, datumReader)) {
    for (SomeAvroClass obj : reader) {
        System.out.println(obj);
    }
}
```

`DataFileStream` constructor calls `void initialize(InputStream in, byte[] magic)` method which extracts `writerSchema` from `inputStream` metadata. At the end the helper method invokes
```
reader.setSchema(header.schema); // replaces writerSchema if it's null --> that's our case
```
where `reader` is our `datumReader`.

Now, in version `0.3.21`, new field was added to `FastGenericDatumReader`:
```
private final FastDeserializer<T> coldDeserializer;
```
which **ignores invocation of `public void setSchema(Schema schema) {`**

Basically **`setSchema(Schema schema)` should be somehow cascaded to `coldDeserializer`.**

----
I did an experiment and injecting 
```
((FastSerdeUtils.FastDeserializerWithAvroSpecificImpl) this.coldDeserializer).customizedDatumReader.setSchema(schema);
```
to `com.linkedin.avro.fastserde.FastGenericDatumReader.setSchema()` fixed the issue.
Of course the real fix should should be done in a better way (i.e. no casting).

FelixGV:
> Basically **`setSchema(Schema schema)` should be somehow cascaded to `coldDeserializer`.**
> 
> I did an experiment and injecting
> 
> ```
> ((FastSerdeUtils.FastDeserializerWithAvroSpecificImpl) this.coldDeserializer).customizedDatumReader.setSchema(schema);
> ```
> 
> to `com.linkedin.avro.fastserde.FastGenericDatumReader.setSchema()` fixed the issue. Of course the real fix should should be done in a better way (i.e. no casting).

Ah, I see, that's very interesting... we should add a test for that (:

krisso-rtb:
Hi @FelixGV @gaojieliu 
I provided a fix for that:
- https://github.com/linkedin/avro-util/pull/534
Please take a look.

We also provided bugfix for another case:
- https://github.com/linkedin/avro-util/pull/538

krisso-rtb:
Hi,
It looks like the feature has a bug which should be fixed by:
- https://github.com/linkedin/avro-util/pull/557

PR #522, opened 2023-11-03
fix complex default conversion - handle list

dg-builder:
`FieldBuilder19`, `FieldBuilder110`, `FieldBuilder111` handle conversion of default values into "acceptable" default values. However, this conversion wasn't handling default values that were non-empty lists.

Added UTs to test this case.

More details on this method in the java doc:
```
  /**
   * we want to be very generous with what we let users provide for default values.
   * sadly, (modern) avro can only handle specific classes/collections/primitive-wrappers
   * (see org.apache.avro.util.internal.JacksonUtils.toJson(Object, JsonGenerator) in 1.9+)
   * @param mightNotBeFriendly a proposed field default value that might originate from
   *                           a call like AvroCompatibilityHelper.getGenericDefaultValue()
   * @return a representation of the input that avro likes for use as a field default value
   */
```


helper/tests/helper-tests-110/src/test/java/com/linkedin/avroutil1/compatibility/avro110/Avro110FieldBuilderTest.java
```diff
@@ -159,6 +160,17 @@ public void testNullDefaultForBoolField() throws Exception {
...
 
+  @Test
+  public void testArrayOfEnumDefaultValue() throws IOException {
+    Schema schema = Schema.parse(TestUtil.load("FieldWithArrayOfEnumDefaultValue.avsc"));
+    Schema.Field field = schema.getField("arrayOfEnum");
+    Object defaultValue = AvroCompatibilityHelper.getGenericDefaultValue(field);
+    FieldBuilder builder = AvroCompatibilityHelper.newField(field);
+    builder.setDefault(defaultValue);
+    Schema.Field resField = builder.build();
+    Assert.assertNotNull(resField.defaultVal());
```
li-ukumar, line 171: Add an assertEquals test as well

dg-builder: Honestly the assert isn't the point of the test. The point of the test is that it doesn't fail to `.build()`. Added a comment in the code

li-ukumar (approved):

PR #524, opened 2023-11-13
initial impl for AvroRecordUtil.validateNestedSchemasConsistent

radai-rosenblatt:


helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroRecordUtil.java
```diff
@@ -811,4 +820,147 @@ private static Enum<?> getSpecificEnumSymbol(Class<? extends Enum<?>> enumClass,
...
+
+  private static void validateNestedSchema(
+          Schema schema,
+          AvroPath path,
+          IdentityHashMap<Schema, Boolean> visited,
+          Map<String, Schema> referenceDefinitions,
+          SchemaComparisonConfiguration comparisonConfiguration
+  ) throws InconsistentSchemaException {
+    if (!visited.containsKey(schema)) {
+      visited.put(schema, Boolean.TRUE); //never again
```
dg-builder, line 951: Can the key here be a fullname string instead of a schema? or do we have to use the schema because of different schemas with the same fullname?

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroRecordUtil.java
```diff
@@ -811,4 +820,147 @@ private static Enum<?> getSpecificEnumSymbol(Class<? extends Enum<?>> enumClass,
...
+  private static void validateNestedRecordSchema(
+          IndexedRecord record,
+          AvroPath path,
+          IdentityHashMap<Schema, Boolean> visited,
+          Map<String, Schema> referenceDefinitions,
+          SchemaComparisonConfiguration comparisonConfiguration
+  ) throws InconsistentSchemaException {
+    Schema recordSchema = record.getSchema();
+
+    //validate current record's schema if we havent before
```
dg-builder, line 859: `if we haven't before`...where's the logic to only validate if we haven't before happening?

dg-builder: Oh it's in the `validateNestedSchema()` call. So we recurse through the whole record but only check the schema if we haven't checked it before

dg-builder (approved):

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroRecordUtil.java
radai-rosenblatt: the issue this is meant to expose is different definitions of the "same" schema. caching on schema name would prevent this from working. this is an identity cache (on instances of org.apache.avro.Schema) for that reason

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/AvroRecordUtil.java
radai-rosenblatt: added doc to the declaration ofthe "visited" map to explain this

PR #526, opened 2023-12-06
codegen: Getters for  collection fields no longer return unmodifiable collection

li-ukumar:
## What
Constructs and returns List and Map views for simple collections, i.e. List<String> and Map<String, String>

## Why
The codegen utilizes List and Map Transformers which return Unmodifiable collections as the field in object are of type Utf8 and return types from getter can be String.
Users might want to modify the object instance directly and make sure it reflects in the original object instance.

## Test
Added Unit tests


helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/collectiontransformer/CollectionTransformerUtil.java
```diff
@@ -17,4 +26,294 @@ public static String getErrorMessageForInstance(Object obj) {
     return String.valueOf(obj) + ((obj == null) ? StringUtils.EMPTY_STRING
         : " (an instance of " + obj.getClass().getName() + ")");
   }
+
+  public static List<String> createStringListView(List<Utf8> utf8List) {
+    return new AbstractList<String>() {
```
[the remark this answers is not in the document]
li-ukumar: Updated.

avro-codegen/src/main/java/com/linkedin/avroutil1/codegen/SpecificRecordGeneratorUtil.java
```diff
@@ -249,6 +249,54 @@ public static boolean isMapTransformerApplicable(AvroSchema schema) {
...
   }
 
+  /**
+   * If a schema is primitive type or a null union of primitive type, it can be handled as a primitive type
+   * @param schema schema to be validated
+   * @return true if schema can be handled as primitive, else false
+   */
+  private static boolean canBeHandledAsPrimitiveType(AvroSchema schema) {
+    if (schema instanceof AvroUnionSchema) {
+      if (((AvroUnionSchema) schema).getTypes().size() != 2) {
```
dg-builder, line 259: if `getTypes().size() == 1`, couldn't it still be handled as a primitive?

avro-codegen/src/main/java/com/linkedin/avroutil1/codegen/SpecificRecordGeneratorUtil.java
```diff
@@ -249,6 +249,54 @@ public static boolean isMapTransformerApplicable(AvroSchema schema) {
...
+        return false;
+      }
+
+      return ((AvroUnionSchema) schema).getTypes().get(0).getSchema().type().equals(AvroType.NULL)
+          && ((AvroUnionSchema) schema).getTypes().get(1).getSchema().type().isPrimitive()
+          || ((AvroUnionSchema) schema).getTypes().get(1).getSchema().type().equals(AvroType.NULL)
+          && ((AvroUnionSchema) schema).getTypes().get(0).getSchema().type().isPrimitive();
```
dg-builder, line 266:
```suggestion
      bool isUnionOfNullAndPrimitive = ((AvroUnionSchema) schema).getTypes().get(0).getSchema().type().equals(AvroType.NULL)
          && ((AvroUnionSchema) schema).getTypes().get(1).getSchema().type().isPrimitive();
      bool isUnionOfPrimitiveAndNull = ((AvroUnionSchema) schema).getTypes().get(1).getSchema().type().equals(AvroType.NULL)
          && ((AvroUnionSchema) schema).getTypes().get(0).getSchema().type().isPrimitive();
      return isUnionOfNullAndPrimitive || isUnionOfPrimitiveAndNull;
```

dg-builder (approved):

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/collectiontransformer/ListTransformer.java
```diff
@@ -60,15 +77,16 @@ private static void flushCharSeqFlag() {
     hasCharSeq.set(false);
   }
 
-  private static List getString(Object listObj) {
+  private static List getString(Object listObj, boolean isPrimitive) {
```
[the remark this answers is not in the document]
li-ukumar: `isPrimitive` is about its values/item which is primitive here. The collection is always complex

avro-codegen/src/main/java/com/linkedin/avroutil1/codegen/SpecificRecordGeneratorUtil.java
li-ukumar: unions of size 1 are not nullable. For example, a union of array is not a nullable array. So we couldn't handle that case

PR #537, opened 2024-01-18
Improve performance of AvscSchemaWriter

karthikrg:
AvscSchemaWriter currently creates a DOM object and then serializes it into a string. This wastes both CPU and memory. Instead use the Jackson streaming API to directly serialize to string. There are APIs available to get a DOM object from a JsonGenerator if needed for debugging purposes.

Add more logging and fix parallelism inside SchemaBuilder.

### DIFF
--- a/avro-builder/builder-spi/src/main/java/com/linkedin/avroutil1/builder/plugins/BuilderPluginContext.java
+++ b/avro-builder/builder-spi/src/main/java/com/linkedin/avroutil1/builder/plugins/BuilderPluginContext.java
@@ -45,19 +45,27 @@ public class BuilderPluginContext {
       throw new IllegalStateException("run() has already been invoked");
     }
 
-    //"seal" any internal state to prevent plugins from trying to do weird things during execution
+    // "seal" any internal state to prevent plugins from trying to do weird things during execution
     sealed = true;
 
-    int operationCount = operations.stream().collect(StreamUtil.toParallelStream(op -> {
-      try {
-        op.run(operationContext);
-      } catch (Exception e) {
-        throw new IllegalStateException("Exception running operation", e);
-      }
+    if (!operations.isEmpty()) {
+      long operationStart = System.currentTimeMillis();
+      final int parallelism = Math.min(operations.size(), 5);
+      int operationCount = operations.stream().collect(StreamUtil.toParallelStream(op -> {
+        try {
+          op.run(operationContext);
+        } catch (Exception e) {
+          throw new IllegalStateException("Exception running operation", e);
+        }
 
-      return 1;
-    }, 2)).reduce(0, Integer::sum);
+        return 1;
+      }, parallelism)).reduce(0, Integer::sum);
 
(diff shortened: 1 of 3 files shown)


avro-builder/builder-spi/src/main/java/com/linkedin/avroutil1/builder/plugins/BuilderPluginContext.java
```diff
@@ -45,19 +45,27 @@ public void run() throws Exception {
...
 
-    int operationCount = operations.stream().collect(StreamUtil.toParallelStream(op -> {
-      try {
-        op.run(operationContext);
-      } catch (Exception e) {
-        throw new IllegalStateException("Exception running operation", e);
-      }
+    if (!operations.isEmpty()) {
+      long operationStart = System.currentTimeMillis();
+      final int parallelism = Math.min(operations.size(), 5);
```
mchen07, line 53: where is this magic number 5 coming from? Also, what is the rationale for previous hard-coded 2?

avro-builder/builder/src/main/java/com/linkedin/avroutil1/builder/SchemaBuilder.java
```diff
@@ -28,20 +28,29 @@
...
 
   public static void main(String[] args) throws Exception {
+    long start = System.currentTimeMillis();
+
+    long pluginLoadStart = System.currentTimeMillis();
     List<BuilderPlugin> plugins = loadPlugins(1);
+    long pluginLoadEnd = System.currentTimeMillis();
+    LOGGER.info("Loaded {} plugins in {} millis.", plugins.size(), pluginLoadEnd - pluginLoadStart);
 
+    long optionParseStart = System.currentTimeMillis();
```
mchen07, line 52: nit: do we need to log options parser? That is pretty standard

avro-builder/builder-spi/src/main/java/com/linkedin/avroutil1/builder/plugins/BuilderPluginContext.java
karthikrg: These plugin operations are generally heavily I/O bound so it is ok to kick off upto 5 threads in parallel. Earlier I hardcoded to 2 because I thought most use cases will have at most 2 plugins, but I realized that avro-schemas itself has 3, so rather than bound it by plugin count, I decided to bound it by hardware overhead.

avro-builder/builder/src/main/java/com/linkedin/avroutil1/builder/SchemaBuilder.java
karthikrg: Plugins can customize option parsing and some of them were doing a bad job perf wise before I cleaned it up. If we have a log, at least we will know if this regresses.

The issue is that since all this runs inside one gradle task, it is hard to look at build scans and figure out where time is being spent. These logs are our only help when debugging perf issues.

PR #355, opened 2022-07-18
Proper handling of asymmetric union reads

flowenol:
Hi

Asymmetric union read operations are not working properly in current avro-fastserde implementation. Such operations are rarely used, but they are clearly defined within Avro [documentation](https://avro.apache.org/docs/current/spec.html#Schema+Resolution) :

**If reader's is a union, but writer's is not**

- The first schema in the reader's union that matches the writer's schema is recursively resolved against it. If none match, an error is signalled.

**if writer's is a union, but reader's is not**

- If the reader's schema matches the selected writer's schema, it is recursively resolved against it. If they do not match, an error is signalled.

The current implementation fails with `FastDeserializerGeneratorException` thrown during deserializer generation every time it encounters asymmetric union reads.  This PR contains fixes to align the behaviour with the one specified in documentation.

Apart from that, the current implementation also can't handle the union as a top level type. This PR also fixes that.

The first commit 33b1b2ee158ddfb275badb143f1fd5f385b5b622 is a bunch of test cases where the current implementation fails.
The second commit 8a9dbd2f4314330e51217cdea1e458e8b57f04cf contains the actual fix.
The third commit 19e0cf2e933c80a3bfb7a18945b891f0a4ae3291 contains the codegen for the fixed test cases.

PR #538, opened 2024-01-18
[fast-avro] Schema evolution: complex fields changed to nullable

kbarczynski-rtbh:
[fast-avro] Fixing case when schema changed and complex fields became nullable.

Scenario: Record is serialized using schema with not-null complex fields (record, map, array)
and later deserialized with changed schema having these fields nullable (union of "null" complex type).


krisso-rtb:
Hi @dg-builder,

It looks like the issue was added more than 1 year ago as part of:
- https://github.com/linkedin/avro-util/pull/355/files#diff-e57cd63db4a7520f3d59cc0e1f67546e95905dd22d978bfbf6c401f06ad07be4

dg-builder (approved):

PR #542, opened 2024-01-19
Add option in SchemaBuilder to ignore user defined json properties during codegen.

li-ukumar:
### What
Adding option to ignore json properties in SchemaBuilder

### Why
During avro codegen, the if the generator detects conflicting defintions of a schema, where the filesystem definition is different from the one in classpath, it throws.
Relaxing this check by allowing the user to specify json properties to be ignored during this check.

It is useful if the duplicate schema is "almost" same but has certain properties which may store metadata for another flow. 

### Test 
Added Sanity Unit test
build

### DIFF
--- a/avro-builder/builder/src/main/java/com/linkedin/avroutil1/builder/SchemaBuilder.java
+++ b/avro-builder/builder/src/main/java/com/linkedin/avroutil1/builder/SchemaBuilder.java
@@ -114,6 +114,11 @@ public class SchemaBuilder {
         .defaultsTo("false")
         .describedAs("true/false");
 
+    OptionSpec<String> jsonPropsToIgnoreInCompareOpt = parser.accepts("jsonPropsToIgnoreInCompare", "json properties to ignore when comparing schemas on filesystem and classpath")
+        .withOptionalArg()
+        .defaultsTo("")
+        .withValuesSeparatedBy(',');
+
     //allow plugins to add CLI options
     for (BuilderPlugin plugin : plugins) {
       plugin.customizeCLI(parser);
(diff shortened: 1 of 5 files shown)


avro-builder/builder/src/main/java/com/linkedin/avroutil1/builder/SchemaBuilder.java
```diff
@@ -238,6 +244,11 @@ public static void main(String[] args) throws Exception {
       skipCodegenIfSchemaOnClasspath = Boolean.TRUE.equals(Boolean.parseBoolean(value));
     }
 
+    List<String> jsonPropsToIgnoreInCompare = new ArrayList<>();
+    if(options.has(jsonPropsToIgnoreInCompareOpt)) {
+      jsonPropsToIgnoreInCompare = options.valuesOf(jsonPropsToIgnoreInCompareOpt);
+    }
+
```
xiaoyu-yang-gh, line 251: nit: change to:

List<String> jsonPropsToIgnoreInCompare = options.has(jsonPropsToIgnoreInCompareOpt) ? options.valuesOf(jsonPropsToIgnoreInCompareOpt) : Collections.emptyList();

avro-builder/builder/src/main/java/com/linkedin/avroutil1/builder/SchemaBuilder.java
```diff
@@ -238,6 +244,11 @@ public static void main(String[] args) throws Exception {
       skipCodegenIfSchemaOnClasspath = Boolean.TRUE.equals(Boolean.parseBoolean(value));
     }
 
+    List<String> jsonPropsToIgnoreInCompare = new ArrayList<>();
```
xiaoyu-yang-gh, line 247: I saw eventually you have this as a HashSet. why not have it as a Set at the beginning?

xiaoyu-yang-gh (approved):

karthikrg:
I don't advise doing this. If folks rely on reading these properties from runtime schema they may get nasty surprises.

PR #546, opened 2024-02-02
Updating default parsing in AvscParser

li-ukumar:
### What
try to get default values from field definitions after schemas in the same file are resolved.

### Why
For a complex schema, with cyclic references, it is possible for field type to be unresolved even though it is fully defined inside the file. So, before checking if we can parse default, wait for all schemas in the file to be resolved.

### Test
Added Unit test.

### DIFF
--- a/parser/src/main/java/com/linkedin/avroutil1/parser/avsc/AvscFileParseContext.java
+++ b/parser/src/main/java/com/linkedin/avroutil1/parser/avsc/AvscFileParseContext.java
@@ -199,9 +199,8 @@ public class AvscFileParseContext {
                 // If schema is not fully defined, add it to the list of fields with unparsed defaults
                 if (!fieldSchema.isFullyDefined()) {
                     fieldsWithUnparsedDefaults.add(field);
-                    continue;
 
-                // schema is defined and default value is unparsed.
+                    // schema is defined and default value is unparsed.
                 } else if (field.getDefaultValue() instanceof AvscUnparsedLiteral) {
                     AvscParseUtil.lateParseFieldDefault(field, this);
                 }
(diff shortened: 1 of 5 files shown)


xiaoyu-yang-gh (commented):
Can you add unit tests?

parser/src/main/java/com/linkedin/avroutil1/parser/avsc/AvscFileParseContext.java
```diff
@@ -178,6 +178,39 @@ public void resolveReferences() {
...
+                    }
+                }
+                //fully defined if we're here
+                if (!hasDefault) {
+                    continue;
+                }
+                if (field.getDefaultValue() instanceof AvscUnparsedLiteral) {
+                    AvscParseUtil.lateParseFieldDefault(field, this);
+                }
+            }
```
xiaoyu-yang-gh, line 212: I think we can refactor these code to remove these two **continue** to make the logic more clear. 

The logic is like:
for (AvroSchemaField field : fields) {
   SchemaOrRef fieldSchema = field.getSchemaOrRef();
   boolean hasDefault = field.hasDefaultValue();
   boolean wasDefinedBefore = fieldSchema.isFullyDefined();  
   if (!wasDefinedBefore) {
       if (hasDefault) {
           fieldsWithUnparsedDefaults.add(field);
       }
   } else {
       if (!hasDefault) {
                if (field.getDefaultValue() instanceof AvscUnparsedLiteral) {
                    AvscParseUtil.lateParseFieldDefault(field, this);
                }
       }
   }
}


Or you can move the the check **if (hasDefault)** ahead:

for (AvroSchemaField field : fields) {
   if (!field.hasDefaultValue()) {
      continue;
   }

   SchemaOrRef fieldSchema = field.getSchemaOrRef();
   boolean wasDefinedBefore = fieldSchema.isFullyDefined();  
   if (!wasDefinedBefore) {
           fieldsWithUnparsedDefaults.add(field);
   } else if (field.getDefaultValue() instanceof AvscUnparsedLiteral) {
           AvscParseUtil.lateParseFieldDefault(field, this);
   }
}

xiaoyu-yang-gh (approved):

PR #548, opened 2024-02-14
fix StreamUtil to execute the map function for each batch in the thread pool

xiaoyu-yang-gh:
methods like **filter** and **map** on stream are lazy. They just create a new stream. Traversal of the pipeline source does not begin until the terminal operation of the pipeline is executed.

The **StreamUtil** wants the **map** method executed for each batch in the thread pool Asynchronously, so we should make the stream pipeline executed in the thread-pool for the reason mentioned above.

Here the collect is the terminal operation that triggers the stream execution.


karthikrg:
LGTM. This is a subtle behavior with flatMap that changed in Java 10. 
https://blog.jooq.org/are-java-8-streams-truly-lazy-not-completely/

karthikrg (approved):

mchen07 (approved):

PR #549, opened 2024-02-15
ColdPrimitive*List classes use correct schema.

krisso-rtb:
Just a minor fix.

The main change is in 4 classes:
- `ColdPrimitiveBooleanList`
- `ColdPrimitiveDoubleList`
- `ColdPrimitiveIntList`
- `ColdPrimitiveLongList`


FelixGV (approved):
Wow, that's a silly mistake! Thanks for catching and fixing it!

PR #550, opened 2024-02-15
optimize StreamUtil and fix the LimitingExecutor

xiaoyu-yang-gh:
(1)  https://github.com/linkedin/avro-util/pull/548 fixed the issue to execute mapper function to each batch in the thread-pool. but the batches are still executed sequentially.  We need to fix the issues to have the batches executed in parallel. Here we implement a Collector:
    - the **accumulate** method collect data from original stream, create batches, and submit jobs for executing each batch to thread-pool.
    - the **finisher** method handles the last batch, submit the job for the last batch to thread-pool. It also connect the result for the jobs executed in the thread-pool, combine the results, and stream the results.
 
Tested this can execute the batches in the thread-pool in parallel.

(2) Fix LimitingExecutor. LimitingExecutor use a Semaphore to control the maximum number of jobs concurrently executed in the thread-pool. But the Semaphore is just used to guard the **ThreadPool.execute(command)**, and the ThreadPool.execute just submit the job to the threadpool. So the Semaphore does not guard the number of jobs concurrently running in the pool. We fix this issue by:
 a. require a Semaphore when submitting a new job;
 b. release a Semaphore when a job completes;

### DIFF
--- a/avro-builder/builder-spi/src/main/java/com/linkedin/avroutil1/builder/util/StreamUtil.java
+++ b/avro-builder/builder-spi/src/main/java/com/linkedin/avroutil1/builder/util/StreamUtil.java
@@ -6,7 +6,11 @@
 
 package com.linkedin.avroutil1.builder.util;
 
-import java.util.Collection;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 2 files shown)


avro-builder/builder-spi/src/main/java/com/linkedin/avroutil1/builder/util/StreamUtil.java
```diff
@@ -72,36 +78,27 @@ private StreamUtil() {
...
+   * @param mapper a transformation to be performed in parallel
+   * @param batchSize the size into which inputs should be batched before running the mapper.
+   * @param executor executor for executing computations in parallel
+   * @param <T> the type of the collected elements
+   * @param <R> the result returned by {@code mapper}
+   *
+   * @return a {@code Collector} which collects all processed elements into a {@code Stream} in parallel.
+   */
+  public static <T, R> Collector<T, ?, Stream<R>> toParallelStream(Function<T, R> mapper, int batchSize,
+      Executor executor) {
```
karthikrg, line 97: Not a fan of exposing this. I think we should provide a sensible default in the util. The reason is that if someone uses an improperly sized pool, it can result in a deadlock.

xiaoyu-yang-gh: Optimized the LimitingExecutor and remove this interface for allowing user pass in executor

xiaoyu-yang-gh:
Close this PR.

PR #555, opened 2024-04-24
Override clear method in custom list views

li-ukumar:
### What
Overriding clear method in List views to run clear methods of provided list rather than defaulting on AbstractList implementation.

### Why
Generic Array implementation in avro has custom implementation of clear, which is lost if AbstractList's method is used.

### Test
build


helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/collectiontransformer/CharSequenceListView.java
```diff
@@ -60,4 +60,9 @@ public boolean addAll(int index, java.util.Collection<? extends CharSequence> c)
   public boolean remove(Object o) {
     return utf8List.remove(new Utf8(o.toString()));
   }
+
+  @Override
+  public void clear() {
+    utf8List.clear();
+  }
 }
```
xiaoyu-yang-gh, line 68: Can you add doc why these method need to be override, the thing you have the PR description.

dg-builder (approved):

dg-builder (approved):

PR #558, opened 2024-06-25
Add for utf8 method in String/Utf8/Charsequence ListViews

li-ukumar:
## What
- Adding add methods for Utf8 type in StringListView

## Test
- Existing unit tests
- build

### DIFF
--- a/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/collectiontransformer/StringListView.java
+++ b/helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/collectiontransformer/StringListView.java
@@ -98,9 +98,4 @@ public class StringListView extends AbstractList<String> {
       }
     };
   }
-
-  @Override
-  public boolean equals(Object o) {
-    return o instanceof StringListView && _utf8List.equals(((StringListView) o)._utf8List);
-  }
 }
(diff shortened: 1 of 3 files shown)


helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/collectiontransformer/StringListView.java
```diff
@@ -38,16 +38,30 @@ public String set(int index, String element) {
...
+
   @Override
   public void add(int index, String element) {
     _utf8List.add(index, new Utf8(element));
   }
 
+  public void add(int index, Utf8 element) {
+    _utf8List.add(index, element);
+  }
+
```
xiaoyu-yang-gh, line 55: no such method in the base class?

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/collectiontransformer/StringListView.java
```diff
@@ -38,16 +38,30 @@ public String set(int index, String element) {
     return previousValue;
   }
 
+  public String set(int index, Utf8 element) {
```
xiaoyu-yang-gh, line 41: should add this method in base class/interface?

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/collectiontransformer/StringListView.java
li-ukumar: No. Base class only has methods for String in this case.. (Since we are extending AbstractList<String>).

helper/helper/src/main/java/com/linkedin/avroutil1/compatibility/collectiontransformer/StringListView.java
li-ukumar: Base class is java's AbstractList. We cant modify that

PR #563, opened 2024-11-15
Support avro 1.11.4.

mjauup:


dg-builder (approved):
thanks @mjauup for the diff!

PR #569, opened 2025-02-28
Fix incorrect check for enableUtf8encoding flag

jairajsolanke:
Fix incorrect check for enableUtf8encoding flag. The incorrect check was preventing from setting/overriding enableUtf8EncodingPutbyIndex flag correctly.


leiykpt (approved):
LGTM

PR #571, opened 2025-02-28
Add support for comparing doc field in avro schemas

SrilakshmiBharadwaj:
## Summary
The current comparator doesn't return false when comparing 2 avro schemas that differ just by doc field. This causes compatibility issues when evolving espresso schemas (adding comment in proto updates avro doc field in the current .avsc version instead of generating a new .avsc)
This PR adds support to compare 2 avro schemas and return false if they differ by doc field

## Testing Done
Added unit tests `testCompareSchemaDoc` and `testCompareSchemaFieldDoc` in `ConfigurableSchemaComparatorTest`

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -29,7 +29,7 @@ jobs:
           java-version: 1.8
       - name: create/restore gradle wrapper cache
         # see inspiration here - https://docs.github.com/en/actions/language-and-framework-guides/building-and-testing-java-with-gradle#caching-dependencies
-        uses: actions/cache@v2
+        uses: actions/cache@v4
         with:
           path: ~/.gradle/wrapper
           key: gradle-wrapper-${{ hashFiles('gradle/wrapper/gradle-wrapper.properties') }}
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -34,7 +34,7 @@ jobs:
           path: ~/.gradle/wrapper
           key: gradle-wrapper-${{ hashFiles('gradle/wrapper/gradle-wrapper.properties') }}
       - name: create/restore gradle cache
-        uses: actions/cache@v2
+        uses: actions/cache@v4
         with:
           path: ~/.gradle/caches
           key: gradle-cache-${{ hashFiles('gradle/wrapper/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle') }}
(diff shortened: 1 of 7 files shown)


.github/workflows/pr.yaml
```diff
@@ -29,12 +29,12 @@ jobs:
           java-version: 1.8
       - name: create/restore gradle wrapper cache
         # see inspiration here - https://docs.github.com/en/actions/language-and-framework-guides/building-and-testing-java-with-gradle#caching-dependencies
-        uses: actions/cache@v2
+        uses: actions/cache@v4
```
li-ukumar, line 32: Why are these changes here?

avro-builder/tests/tests-allavro/src/test/java/com/linkedin/avroutil1/builder/SpecificRecordTest.java
```diff
@@ -118,8 +118,8 @@ private Object[][] TestRoundTripSerializationProvider() {
 
         {charseqmethod.TestCollections.class, charseqmethod.TestCollections.getClassSchema()},
```
li-ukumar, line 119: since the following lines are commented out, shouldn't the `,` be removed from this line too?

.github/workflows/pr.yaml
SrilakshmiBharadwaj: Build signals were failing and complaining v2 is out of date, need to upgrade to v3/v4

helper/helper-common/src/main/java/com/linkedin/avroutil1/compatibility/SchemaComparisonConfiguration.java
```diff
@@ -15,16 +15,16 @@ public class SchemaComparisonConfiguration {
...
   /**
    * same as {@link #STRICT} but allows int default values to match (round) float default values
    */
   public static final SchemaComparisonConfiguration LOOSE_NUMERICS = new SchemaComparisonConfiguration(
-      true, true, true, true, true, true, Collections.emptySet()
+      true, true, true, true, true, true, false, Collections.emptySet()
   );
   public static final SchemaComparisonConfiguration STRICT = new SchemaComparisonConfiguration(
-      true, true, true, false, true, true,  Collections.emptySet()
+      true, true, true, false, true, true,  true, Collections.emptySet()
```
li-ukumar, line 27: This would start failing existing users of this configuration.. Lets create a separate config with this check turned on.. something like `STRICT_COMPARE_FIELD_DOCS` ?

PR #574, opened 2025-04-03
Compatible Record In Union Lookup Should Use Unqualified Name

pamitm:
The avro 1.8 and before spec says that two records match if their names are the same. However, more recent versions like 1.11 clarify that this matching should be done based on the unqualified name. The unit test added here and our experience with 1.8 at Uber suggest that all Avro versions except 1.5-1.7 use the unqualified name for the actual resolution. Thus, the behavior of the current compatible schema in union finder deviates from the standard avro impl and this change brings the two back in sync.

### DIFF
--- a/fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
+++ b/fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
@@ -663,10 +663,7 @@ public class FastDeserializerGenerator<T, U extends GenericData> extends FastDes
             Schema potentialReaderSchema = unionReaderSchema.getTypes().get(j);
             // Avro allows unnamed types to appear only once in a union, but named types may appear multiple times and
             // thus need to be disambiguated via their full-name (including aliases).
-            if (potentialReaderSchema.getType().equals(optionSchema.getType()) &&
-                (!schemaAssistant.isNamedType(potentialReaderSchema) ||
-                    AvroCompatibilityHelper.getSchemaFullName(potentialReaderSchema).equals(AvroCompatibilityHelper.getSchemaFullName(optionSchema)) ||
-                    potentialReaderSchema.getAliases().contains(AvroCompatibilityHelper.getSchemaFullName(optionSchema)))) {
+            if (schemaAssistant.areTypesCompatible(potentialReaderSchema, optionSchema)) {
               readerOptionSchema = potentialReaderSchema;
               readerOptionUnionBranchIndex = j;
               break;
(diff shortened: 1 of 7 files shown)


pamitm:
I have no idea how the error even happened but just rebasing and pushing seems to have fixed it.

pamitm:
@gaojieliu could you take another look at this? Thanks!

gaojieliu:
@pamitm 
Sorry for the late review.
Can you also checkin the generated classes otherwise, the release will be blocked?

pamitm:
@gaojieliu added. Thanks!

pamitm:
I think we have a flaky test on our hands here. It's failed again for no understandable reason. I'm going to comment it out if that's ok @gaojieliu

pamitm:
Another PR also had the same failure https://github.com/linkedin/avro-util/actions/runs/15353472419/job/43207087020

gaojieliu:
@pamitm 
So this test flakiness happens even without your recent code changes, right?

pamitm:
I think so but it's hard to confirm since I've been unable to reproduce the issue locally and all the old failing job logs have expired. I don't see how it could be connected to any of the changes I've made.

gaojieliu:
@pamitm 
Another PR fixed the test flakiness:
https://github.com/linkedin/avro-util/pull/584
So can you rebase and update the PR?

pamitm:
@gaojieliu Thanks! Rebased. Please take another look.

gaojieliu:
@pamitm 
The latest release is out:
https://linkedin.jfrog.io/ui/native/avro-util/com/linkedin/avroutil1/avro-fastserde/0.4.32/

fastserde/avro-fastserde/src/main/java/com/linkedin/avro/fastserde/FastDeserializerGenerator.java
```diff
@@ -663,10 +663,7 @@ private void processUnion(JVar unionSchemaVar, final String name, final Schema u
             Schema potentialReaderSchema = unionReaderSchema.getTypes().get(j);
             // Avro allows unnamed types to appear only once in a union, but named types may appear multiple times and
             // thus need to be disambiguated via their full-name (including aliases).
-            if (potentialReaderSchema.getType().equals(optionSchema.getType()) &&
-                (!schemaAssistant.isNamedType(potentialReaderSchema) ||
-                    AvroCompatibilityHelper.getSchemaFullName(potentialReaderSchema).equals(AvroCompatibilityHelper.getSchemaFullName(optionSchema)) ||
-                    potentialReaderSchema.getAliases().contains(AvroCompatibilityHelper.getSchemaFullName(optionSchema)))) {
+            if (schemaAssistant.areTypesCompatible(potentialReaderSchema, optionSchema)) {
```
maciejkowalczyk, line 666: Hi @pamitm!
I've noticed that arguments are swapped in this line.

PR #577, opened 2025-04-16
Support java keywords as class names for records during generic decode

pamitm:


gaojieliu:
Hi @pamitm 
Can you share the generated code for your new test case before/after this change?

pamitm:
Attached before and after.
[reservedNameCodegenAfter.txt](https://github.com/user-attachments/files/19968642/reservedNameCodegenAfter.txt)
[reservedNameCodegenBefore.txt](https://github.com/user-attachments/files/19968643/reservedNameCodegenBefore.txt)

PR #576, opened 2025-04-15
Fix nested map and array problems

pamitm:
The root cause of the issue here is that when passing the actions to a record, we are not passing in the relevant actions. Rather, the processRecord has been written to take in symbols from arbitrary levels preceding the record symbol and then seek to the correct action inside that. This diff tries to ameliorate this problem slightly by making the logic for maps and arrays consistent. 

The comment in processMap covers the idea of how we now pass the symbols to the the elements and value types for arrays and maps respectively. 

The proper fix for this would be to do this to change the traversal of the symbol tree and visit each symbol just once as well as assert the invariant that we visit all of them. 

Fixes #506 as well as the related problem for nested arrays not mentioned in the original ticket

PR #578, opened 2025-04-28
Fix the map/array for nested maps/array with unions at play

pamitm:
Missed a few cases in #576 when we have nested maps or array with the following cases:
1. In case we have a map<map<int>> inside a union, the old logic was passing a production which had the map-start,map-end symbols. However while processing maps our new map logic expects to see a map-end symbol as the action which we then check the production for to find the actions symbols for the map value. In this PR, we change union processing to seek to this action before passing to the map. Once again this indicates that we need a rewrite of this entire action processing logic to standardize which symbols are expected when we process any type to avoid unnecessary seeks and reduce confusion. 
2. If we have map<map<int>> on writer and map<union<map<int>> on reader there's actually no ambiguity about which map-end symbol to use for the value action since the inner map symbol is now contained in a UnionAdjustAction. We account for this by modifying our checks in processMap and processArray

We tested against ~30,000 schemas at Uber and this new logic looks sound 🤞

### DIFF
--- a/fastserde/avro-fastserde-tests-common/src/test/java/com/linkedin/avro/fastserde/FastGenericDeserializerGeneratorTest.java
+++ b/fastserde/avro-fastserde-tests-common/src/test/java/com/linkedin/avro/fastserde/FastGenericDeserializerGeneratorTest.java
@@ -1734,6 +1734,181 @@ public class FastGenericDeserializerGeneratorTest {
     Assert.assertEquals(outerRecord2.toString(), outerRecord1.toString());
   }
 
+  @Test(groups = {"deserializationTest"}, dataProvider = "Implementation")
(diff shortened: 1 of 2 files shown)


gaojieliu (approved):

gaojieliu:
@pamitm 
The new release is out: 0.4.26

gaojieliu:
@pamitm 
Actually, we found some issues with the release procedure, and I will update here once we fix it.

gaojieliu:
@pamitm 
please try 0.4.30:
https://linkedin.jfrog.io/ui/native/avro-util/com/linkedin/avroutil1/avro-fastserde/0.4.30/

pamitm:
Thank you so much Gaojie!

PR #585, opened 2025-07-17
Replace remaining occurrences of example.com URL

sandaar:
This is a follow-up PR for https://github.com/linkedin/avro-util/pull/584/  to replace remaining occurrences of example.com URL (see the [search query](https://github.com/search?q=repo%3Alinkedin%2Favro-util+%22http%3A%2F%2Fwww.example%22&type=code)). We want to use un-resolvable URLs in tests to avoid intermittent test failures due to URLs resolving to different IPs.


gaojieliu (approved):
LGTM, thanks!

PR #591, opened 2025-09-22
Support for selectively demoting long to int in avro 1.5-1.8

karthikrg:
Support for selectively demoting long to int in avro 1.5-1.8


helper/impls/helper-impl-15/src/main/java/com/linkedin/avroutil1/compatibility/avro15/backports/GenericDatumReaderExt.java
```diff
@@ -18,7 +23,123 @@
...
+    super.setSchema(writer);
+    this.writer = writer;
+    if (reader == null) {
+      this.reader = writer;
+    }
+  }
+
+  @SuppressWarnings("unchecked")
+  @Override
+  public T read(T reuse, Decoder in) throws IOException {
```
li-ukumar, line 61: javadocs on the public method please.

helper/impls/helper-impl-16/src/main/java/com/linkedin/avroutil1/compatibility/avro16/parsing/ResolvingGrammarGenerator.java
```diff
@@ -171,6 +171,9 @@ public Symbol generate(
       case NULL:
       case BOOLEAN:
       case INT:
+        if (writerType == Schema.Type.LONG) {
+          return Symbol.IntLongAdjustAction.INSTANCE;
+        }
```
bipinparmar, line 177: What is the reason to change the behavior for writer schema = long and reader schema = boolean. Is it because we have checks in schema evolution that Boolean datatype cannot be changed to Long and vice-versa. It might be better to put a "break;" right below "case BOOLEAN:"

sandaar (approved):
LGTM!

